PR #2, opened 2019-06-12
Br/convo styling

breitman:


hallettj:
I'm seeing some lint errors:

```
./src/Conversation.tsx
  Line 250:  'PresentableEdited' is defined but never used          @typescript-eslint/no-unused-vars
  Line 259:   'editorIsntAuthor' is assigned a value but never used  @typescript-eslint/no-unused-vars
```

You can see errors in the terminal where you run the `yarn start` client process.

packages/client/src/Conversation.tsx
```diff
@@ -159,42 +170,80 @@ function Presentable({
...
-            <>
-              <button
+    <Card className={classes.presentable}>
+      <CardHeader
+        title={displayParticipant(presentable.from)}
+        avatar={<Avatar address={presentable.from} />}
+        subheader={
+          moment(presentable.date).calendar() +
+          " " +
+          displayPresentableEdited(presentable)
```
hallettj, line 193: We don't want to display an edited message if the content has not been edited. Messages that have not been edited are showing "null" here.

packages/client/src/Conversation.tsx
```diff
@@ -159,42 +170,80 @@ function Presentable({
...
+
+  function handleClick(event: React.MouseEvent<HTMLButtonElement>) {
+    setAnchorEl(event.currentTarget)
+  }
+
+  function handleClose() {
+    setAnchorEl(null)
+  }
+
+  //TODO: set aria-control's id
```
hallettj: Let's go ahead and set up that aria-control id.

packages/client/src/Conversation.tsx
```diff
@@ -159,42 +170,80 @@ function Presentable({
...
-          )
+              </MenuItem>
+            </Menu>
+          </div>
         }
-      })}
-    </div>
+        // <strong>
+        //   <Participant {...presentable.from} />
+        // </strong>{" "}
```
hallettj: Please delete code that has been commented out.

packages/client/src/Conversation.tsx
```diff
@@ -212,17 +261,27 @@ function PresentableEdited({
...
-      {editorIsntAuthor ? (
-        <>
-          {" "}
-          by <Participant {...presentable.editedBy} />
-        </>
-      ) : null}
+      {displayPresentableEdited(presentable)}
     </span>
   )
 }
```
hallettj: `PresentableEdited` is not used anymore, so it can be removed.

packages/client/src/Conversation.tsx
```diff
@@ -212,17 +261,27 @@ function PresentableEdited({
...
 
+function displayPresentableEdited({
+  editedAt,
+  editedBy,
+  from
+}: graphql.Presentable) {
+  //const classes = useStyles()
```
hallettj: Please delete commented code.

packages/client/src/Conversation.tsx
```diff
@@ -212,17 +261,27 @@ function PresentableEdited({
...
+  if (!editedAt || !editedBy) {
+    return null
+  }
+  const editorIsntAuthor =
+    editedBy.host !== from.host || editedBy.mailbox !== from.mailbox
+  return `Edited ${moment(editedAt).calendar()} ${
+    editorIsntAuthor ? ` by ${<Participant {...editedBy} />}` : ""
```
hallettj: This is nice.

hallettj (changes requested):

packages/client/src/Conversation.tsx
hallettj: Oh, I overlooked something: this function returns strings, so you cannot call the `Participant` react component. When I view an edited conversation with this code this is what I see:

> Last Friday at 11:52 AM Edited Last Friday at 12:00 PM by [object Object]

React elements and strings do not mix. The only thing you can put inside a string is another string.

hallettj:
This is good progress. There are just two issues to address: change `displayPresentableEdited` so that it does not call a React component, and hook up the aria-control ID for the menu button.

hallettj (approved):
Looking good! I made changes to remove the "TODO" comment, remove an unused import, and to alphabetize the import list. When you are ready I would like you to "squash and merge" these changes. Please write a commit message in the format described here:

https://gist.github.com/stephenparish/9941e89d80e2bc58a153

Using a consistent format for commit messages in the master branch makes it possible to automatically generate change logs. This change is a feature, so the commit message might look like this:

> feat: display messages in cards

PR #3, opened 2019-06-12
Br/delete accounts

breitman:
- made that open close button a '+' or 'x' depending if open or close and made it so you can close the modal by clicking on 'x' button. 
- I couldn't figure out why when pressing the trash icon it wouldn't delete the account, so that isn't fully functional
-I also couldn't figure out how to make the cards the same size


packages/client/src/Accounts.tsx
```diff
@@ -1,15 +1,87 @@
 import { Link, RouteComponentProps } from "@reach/router"
 import * as React from "react"
 import * as graphql from "./generated/graphql"
+import { createStyles, makeStyles, Theme } from "@material-ui/core/styles"
```
hallettj: There are a few unused imports here.

packages/client/src/Accounts.tsx
```diff
@@ -40,37 +112,84 @@ export default function Accounts(_props: RouteComponentProps) {
...
+      <div className={classes.appBarSpacer} />
+      <AppBar>
+        <Toolbar>
+          <Typography variant="h6">Poodle</Typography>
+        </Toolbar>
+      </AppBar>
+      <div className={classes.root}>
+        {accounts.map(account => (
+          <section key={account.id}>
+            <header>
```
hallettj: This `<header>` tag is not necessary. Usually a `<header>` is contains a heading, such as an `<h1>` tag.

packages/client/src/Accounts.tsx
```diff
@@ -40,37 +112,84 @@ export default function Accounts(_props: RouteComponentProps) {
...
+                <CardHeader title={account.email} />
+                <CardActions>
+                  {account.loggedIn ? (
+                    <>
+                      <Link
+                        to={`/accounts/${account.id}/dashboard`}
+                        className={classes.link}
+                      >
+                        <Button>View Conversations</Button>
+                      </Link>
```
hallettj: This looks like it works; but the idiomatic way to combine a `Button` with a custom link component is to put the link in the button's `component` property. I think you want something like this:

```typescript
<Button component={Link} to={`/accounts/${account.id}/dashboard`}>View Conversations</Button>
```

By default the `Button` component will render a native HTML `<a>` or `<button>` tag - but if you specify `component={Link}` it will render a `<Link>` element instead, and will pass the `to` prop forward to that `Link`.

packages/client/src/Accounts.tsx
```diff
@@ -1,15 +1,87 @@
...
+    justifyContent: "spaceAround",
+    alignContent: "spaceAround"
+  },
+  card: {
+    marginTop: theme.spacing(2),
+    marginLeft: theme.spacing(1),
+    marginRight: theme.spacing(1),
+    height: "100%"
+  },
+  link: {
```
hallettj: I think that this `link` class is not necessary.

packages/client/src/Accounts.tsx
```diff
@@ -1,15 +1,87 @@
...
+    flexDirection: "row",
+    flexWrap: "wrap",
+    justifyContent: "spaceAround",
+    alignContent: "spaceAround"
+  },
+  card: {
+    marginTop: theme.spacing(2),
+    marginLeft: theme.spacing(1),
+    marginRight: theme.spacing(1),
+    height: "100%"
```
hallettj: AFAICT this `height` rule is not necessary.

packages/client/src/Accounts.tsx
```diff
@@ -1,15 +1,87 @@
...
+    bottom: theme.spacing(2),
+    zIndex: theme.zIndex.snackbar
+  },
+  appBarSpacer: theme.mixins.toolbar as any,
+  root: {
+    display: "flex",
+    flexDirection: "row",
+    flexWrap: "wrap",
+    justifyContent: "spaceAround",
+    alignContent: "spaceAround"
```
hallettj: When writing this CSS-in-JS the rule names are camel-cased, but the values have dashes as usual. The values here should be `"space-around"`, not `"spaceAround"`. If you were having trouble getting the layout to work properly that might be why.

packages/client/src/Accounts.tsx
```diff
@@ -40,37 +112,84 @@ export default function Accounts(_props: RouteComponentProps) {
...
+                        className={classes.link}
+                      >
+                        <Button>View Conversations</Button>
+                      </Link>
+                      <IconButton
+                        aria-label="Delete"
+                        onClick={() =>
+                          deleteAccount({
+                            variables: { id: account.id }
+                          }).catch(setError)
```
hallettj: Calling `setError` - very nice!

packages/client/src/Accounts.tsx
```diff
@@ -40,37 +112,84 @@ export default function Accounts(_props: RouteComponentProps) {
...
+        color="secondary"
+        aria-label="Add Acount"
+        className={classes.fab}
+        onClick={handleClick}
+      >
+        {open ? <CloseIcon /> : <AddIcon />}
+      </Fab>
+      <Modal
+        aria-labelledby="add account modal"
+        aria-describedby="add email account"
```
hallettj: The values for `aria-labelledby` and `aria-describedby` should be DOM ids. `labelledby` should match an id on the modal's title element, and `describedby` should match an id on the descriptive content.

packages/main/src/resolvers/account.test.ts
```diff
@@ -319,6 +319,59 @@ describe("sync", () => {
       expect.any(ConnectionManager)
     )
   })
+  it("deletes an account", async () => {
```
hallettj, line 323: Sorry to ask for such a trivial change, but I would like to have blank lines between tests.

packages/main/src/resolvers/account.ts
```diff
@@ -61,6 +61,11 @@ export const AccountMutations: AccountMutationsResolvers = {
     }
     await schedule(actions.sync({ accountId: id }))
     return account
+  },
+  delete(_parent, { id }) {
```
hallettj, line 66: I would also like blank lines between method implementations.

packages/client/src/Accounts.tsx
```diff
@@ -40,37 +112,84 @@ export default function Accounts(_props: RouteComponentProps) {
...
+          </section>
+        ))}
+      </div>
+      <Fab
+        color="secondary"
+        aria-label="Add Acount"
+        className={classes.fab}
+        onClick={handleClick}
+      >
+        {open ? <CloseIcon /> : <AddIcon />}
```
hallettj, line 165: This pop-up form is looking good! Thank you for putting in the icon swap.

packages/client/src/Accounts.tsx
```diff
@@ -1,15 +1,87 @@
...
+    boxShadow: theme.shadows[5],
+    padding: theme.spacing(4),
+    outline: "none"
+  },
+  addAcountButton: {
+    top: theme.spacing(2)
+  }
+}))
+
+//TODO: add functionality to delete button
```
hallettj: Please delete this "TODO".

hallettj (changes requested):
This is great progress! The delete button works on my machine. I did have to restart my main process to pick up the graphql schema changes and the new resolver.

hallettj:
I think there might be some more unused imports. VS Code has an "organize imports" feature that will remove unused imports, and sort everything in a consistent order. It is a good habit to run that, and you can opt into organize imports every time you save with these instructions: https://stackoverflow.com/a/55968714/103017

Please also remove commented-out code.

After those changes this will be ready to merge. Please squash & merge with a commit message based on the Angular commit conventions as you did yesterday.

Issue #4, opened 2019-06-13
autocomplete recipient addresses


hallettj:
Acceptance checklist:

- [x] Results should be consistently ordered (e.g. `order by name, mailbox, host`)
- [x] Test that we only get a single result per email address when the same person is listed multiple times in the `message_participants` table (e.g. `group by mailbox, host`)
- [x] Test a case where participant `name` is `null`

You should write these tests for the contact sync module:

- [x] after calling `downloadContacts` data from the fake API response should be in the local database
- [x] test a case where the fake response data includes a name, and one where it doesn't
- [x] when running `downloadContacts` twice the second API request should include the sync token from the response to the first request

You will probably need to update `sync.test.ts` with the same mocks that you set up in `contacts.test.ts`. Make sure that all tests pass.

PR #6, opened 2019-06-18
auto-completion feature fixes 

breitman:
full-width for the "To" text-field still not working, may have to do with inherited position?


hallettj:
This pull request addresses an issue that I assigned to you to track the autocomplete feature. I'd like you to link pull requests to relevant issues in cases like this, so that the issue automatically closes when the pull request is merged. You can do that ~with a comment like this~ by putting a note like this in the body of the pull request or in the squashed & merged commit message:

resolves #4

packages/client/package.json
```diff
@@ -27,6 +30,8 @@
     "marked": "^0.6.2",
     "react": "^16.8.6",
     "react-apollo": "^3.0.0-beta.1",
+    "react-apollo-hooks": "^0.4.5",
```
hallettj: The `react-apollo-hooks` dependency should not be here. I think that probably got pulled in with the merge conflicts.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -27,10 +56,15 @@ type Action =
 
 type State = { recipients: Address[]; inputValue: string }
 
+interface Suggestion {
```
hallettj: This interface is not used anymore.

packages/client/src/Compose/index.tsx
```diff
@@ -66,6 +66,7 @@ const useStyles = makeStyles(theme => ({
     justifyContent: "stretch"
   },
   formInput: {
+    width: "100%",
```
hallettj: Does setting width here change anything?

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -73,40 +107,124 @@ export default function RecipientsInput({ onRecipients, ...rest }: Props) {
...
+  const [stateSuggestions, setSuggestions] = React.useState<graphql.Address[]>(
+    []
+  )
+
   const [focused, setFocused] = React.useState(false)
 
+  const { data } = graphql.useGetAllAddressesQuery({
+    variables: { inputValue }
+  })
+  let addresses = data && data.addresses
```
hallettj: Use `const` for variables that are not going to be reassigned.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -73,40 +107,124 @@ export default function RecipientsInput({ onRecipients, ...rest }: Props) {
...
+
+  const suggestions = addresses || []
```
hallettj: It is not necessary to introduce a new variable since you already have `addresses`. You could change the initialization of `addresses` to set it to an empty array if data is not loaded, and get rid of `suggestions`.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -73,40 +107,124 @@ export default function RecipientsInput({ onRecipients, ...rest }: Props) {
...
+
+  function renderSuggestion(
```
hallettj, line 131: This rendering is very nice! I like that matching portions of text are highlighted.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -73,40 +107,124 @@ export default function RecipientsInput({ onRecipients, ...rest }: Props) {
...
+            >
+              {result.text}
+            </span>
+          ))}
+        </div>
+      </MenuItem>
+    )
+  }
+
+  function getSuggestions() {
```
hallettj: `getSuggestions` does not do anything. You can remove this function, and refer to the `suggestions` variable directly.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -73,40 +107,124 @@ export default function RecipientsInput({ onRecipients, ...rest }: Props) {
     recipients: [],
     inputValue: ""
   })
+  const [stateSuggestions, setSuggestions] = React.useState<graphql.Address[]>(
+    []
+  )
+
   const [focused, setFocused] = React.useState(false)
 
+  const { data } = graphql.useGetAllAddressesQuery({
```
hallettj: I would like to skip running this query if `inputValue` has fewer than three characters after whitespace has been trimmed. You can use the `skip` parameter described in the [react-apollo docs](https://www.apollographql.com/docs/react/essentials/queries/).

packages/client/src/documents/operations.graphql
```diff
@@ -93,6 +93,14 @@ query getConversation($id: ID!) {
   }
 }
 
+query getAllAddresses($inputValue: String!) {
```
hallettj: The name of this query is a little misleading. I think something like "getMatchingAddresses" might be more descriptive.

packages/main/src/resolvers/index.test.ts
```diff
@@ -0,0 +1,195 @@
...
+    null,
+    null,
+    { value: "bob" }
+  )
+  expect(result).toEqual({
+    data: {
+      addresses: [{ name: null, host: "test.com", mailbox: "bob" }]
+    }
+  })
+})
```
hallettj, line 195: nice tests!

hallettj (changes requested):
There are a couple more issues to resolve. Pressing <kbd>Enter</kbd> should accept a suggestion, but instead it sends the message. And I can see the label of the next input through the suggestions list.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -32,7 +33,7 @@ const useStyles = makeStyles(theme => ({
   },
   suggestionsContainerOpen: {
     position: "absolute",
-    zIndex: 1,
+    zIndex: 6,
```
hallettj: The Material UI theme provides some predefined z-index values. We should use one of them, such as `theme.zIndex.snackbar`, instead of a hard-coded number.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -143,22 +146,18 @@ export default function RecipientsInput({ onRecipients, ...rest }: Props) {
...
   }
 
   const handleSuggestionsFetchRequested = () => {
-    setSuggestions(getSuggestions())
+    return suggestions
   }
 
   const handleSuggestionsClearRequested = () => {
-    setSuggestions([])
+    suggestions = []
```
hallettj: This assignment will not work because the change will not persist across re-renders. It is also unnecessary because of the `skip ? [] : suggestions` check below. I recommend removing `handleSuggestionsClearRequested` entirely.

packages/client/src/Compose/index.tsx
```diff
@@ -170,6 +175,7 @@ export default function Compose({ accountId }: Props) {
             Send
           </Button>
         </form>
+        <script></script>
```
hallettj: Please remove this script tag

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -67,46 +101,130 @@ function reducer(state: State, action: Action): State {
...
+
   const [focused, setFocused] = React.useState(false)
 
+  const { data } = graphql.useGetMatchingAddressesQuery({
+    variables: { inputValue }
+  })
+
+  let suggestions = data ? data.addresses : []
+
+  const skip = inputValue.trim().length < 3
```
hallettj: This is good! But you should also pass `skip` to `useGetMatchingAddressesQuery` so that the client does not send a request when you do not want completions.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -67,46 +101,130 @@ function reducer(state: State, action: Action): State {
...
+  ) {
+    const matches = match(getSuggestionValue(suggestion), query)
+    const results = parse(getSuggestionValue(suggestion), matches)
+
+    return (
+      <MenuItem selected={isHighlighted} component="div">
+        <div>
+          {results.map(result => (
+            <span
+              key={result.text}
```
hallettj: This is a case where it would be better to use `index` for the key because it is possible that there will be duplicate `text` values, and using the same key for multiple elements in the same list is problematic.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -67,46 +101,130 @@ function reducer(state: State, action: Action): State {
...
+    )
+  }
+
+  function getSuggestionValue(suggestion: graphql.Address) {
+    return suggestion.name
+      ? `${suggestion.name} <${suggestion.mailbox}@${suggestion.host}>`
+      : `${suggestion.mailbox}@${suggestion.host}`
+  }
+
+  const handleSuggestionsFetchRequested = () => {
```
hallettj: This callback is not doing anything. Its purpose is to signal that a fetch should be made to get updated suggestions. But `useGetMatchingAddressesQuery` automatically fetches suggestions whenever the input value changes.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -67,46 +101,130 @@ function reducer(state: State, action: Action): State {
...
+    renderInputComponent,
+    suggestions: skip ? [] : suggestions,
+    onSuggestionsFetchRequested: handleSuggestionsFetchRequested,
+    onSuggestionsClearRequested: handleSuggestionsClearRequested,
+    onSuggestionSelected,
+    getSuggestionValue,
+    renderSuggestion
+  }
+
   React.useEffect(() => {
```
hallettj: Please group hooks at the top of the component. It is important that every hook runs every time the component renders; if a hook is called too far down we might accidentally put in an early return or something that could break things. I also like having hooks at the top because they inform someone reading the code of important side-effects, and it is useful to be able to look at those details in one place.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -67,46 +101,130 @@ function reducer(state: State, action: Action): State {
...
-              />
-            ))}
-          </InputAdornment>
-        ),
-        onChange: event =>
-          dispatch({ type: "inputChange", value: event.target.value }),
+    <Autosuggest
+      {...autosuggestProps}
+      inputProps={{
+        id: "react-autosuggest-simple",
```
hallettj: I think that this `id` is not necessary. The documentation says, "Use it only if you have multiple Autosuggest components on a page."

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -67,46 +101,130 @@ function reducer(state: State, action: Action): State {
...
     inputValue: ""
   })
+
   const [focused, setFocused] = React.useState(false)
 
+  const { data } = graphql.useGetMatchingAddressesQuery({
+    variables: { inputValue }
+  })
+
+  let suggestions = data ? data.addresses : []
```
hallettj: I would still like you to change this declaration to `const suggestions` because this variable should not be reassigned.

packages/client/src/Compose/index.tsx
```diff
@@ -137,7 +137,13 @@ export default function Compose({ accountId }: Props) {
       <main className={classes.content}>
         <div className={classes.appBarSpacer} />
         <DisplayErrors results={[sendMessageResult]} />
-        <form className={classes.form} onSubmit={handleSubmit}>
+        <form
+          className={classes.form}
+          onSubmit={handleSubmit}
+          onKeyPress={e => {
+            e.key === "Enter" && e.preventDefault()
```
hallettj, line 144: Looks good!

packages/main/schema/schema.graphql
```diff
@@ -105,6 +105,7 @@ type Mutation {
 type Query {
   account(id: ID!): Account
   accounts: [Account!]!
+  addresses(inputValue: String!): [Address!]!
```
hallettj, line 108: Good stuff

hallettj (changes requested):
The progress is looking good! I added a lot of comments. Please do not take as a reflection on your work - this is a complicated feature so there are a lot of details to attend to.

hallettj:
It looks like there are changes to `yarn.lock` that still need to be committed.

packages/client/src/Compose/RecipientsInput.tsx
```diff
@@ -121,11 +121,14 @@ export default function RecipientsInput({
...
   })
+
   React.useEffect(() => {
     onRecipients(recipients)
   }, [onRecipients, recipients])
 
-  const suggestions = data ? data.addresses : []
+  console.log(data)
+
+  const suggestions = data && data.addresses ? data.addresses : []
```
hallettj, line 129: Nice fix!

packages/main/src/sync.ts
```diff
@@ -31,17 +31,17 @@ export async function sync(accountId: cache.ID, manager: ConnectionManager) {
...
-      box,
-      boxId,
-      manager
-    }).sync()
+    // await new BoxSync({
+    //   accountId,
+    //   box,
+    //   boxId,
+    //   manager
+    // }).sync()
```
hallettj: Please re-enable this `BoxSync` call.

packages/main/src/sync.ts
```diff
@@ -31,17 +31,17 @@ export async function sync(accountId: cache.ID, manager: ConnectionManager) {
...
   }
   const contactApiClient = AccountManager.getContactsApiClient(
     String(accountId)
   )
-  contactApiClient && contactApiClient.downloadContacts(accountId)
+  contactApiClient && (await contactApiClient.downloadContacts(accountId))
```
hallettj, line 44: Good idea to put the `await` here. For one thing if there is an error in the `downloadContacts` call the error will now propagate to the result of `sync`. Without this `await` the error would be lost.

Later we can talk about how to run asynchronous calls in parallel.

packages/main/src/contacts.ts
```diff
@@ -19,6 +19,7 @@ export default class ContactsApiClient {
   }
 
   async getContacts(accountId: ID) {
+    console.log(google.getSupportedAPIs())
```
hallettj: Please remove this `console.log`

hallettj (changes requested):
Just a couple of items left to fix:
- restore commented code in `sync.ts`
- remove a `console.log` call
- remove the `handleSuggestionsFetchRequested` which is not doing anything

PR #7, opened 2019-06-18
closing IMAP connection upon deleting account

### LINKED ISSUE #5: remove connections in AccountManager when an account is deleted
While an account is logged in Poodle keeps a connection open to account's IMAP server. When the user deletes an account or logs out that connection should be cleaned up.

breitman:
resolves #5

### DIFF
--- a/packages/main/src/managers/AccountManager.ts
+++ b/packages/main/src/managers/AccountManager.ts
@@ -69,7 +69,7 @@ export default class AccountManager {
 
   static deleteConnectionManager(accountId: ID) {
     const connection = this.getConnectionManager(accountId)
-    connection ? connection.closeConn() : null
+    connection && connection.closeConn()
 
     delete this.connectionManagers[accountId]
   }
--- a/packages/main/src/managers/AccountManager.ts
+++ b/packages/main/src/managers/AccountManager.ts
@@ -67,6 +67,13 @@ export default class AccountManager {
     return this.connectionManagers[accountId]
   }
 
+  static deleteConnectionManager(accountId: ID) {
+    const connection = this.getConnectionManager(accountId)
+    connection ? connection.closeConn() : null
+
+    delete this.connectionManagers[accountId]
+  }
+
   static async addSmtpTransporter(
     account: Account,
     credentials: OAuthCredentials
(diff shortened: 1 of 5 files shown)


packages/main/src/managers/AccountManager.ts
```diff
@@ -67,6 +67,13 @@ export default class AccountManager {
     return this.connectionManagers[accountId]
   }
 
+  static deleteConnectionManager(accountId: ID) {
```
hallettj, line 70: perfect!

packages/main/src/managers/ConnectionManager.ts
```diff
@@ -53,7 +54,7 @@ export default class ConnectionManager {
           if (this.conn === connPromise) {
             this.conn = null
           }
-          if (this.options.keepalive) {
+          if (this.options.keepalive && !this.isClose) {
```
hallettj, line 57: Great!

packages/main/src/managers/ConnectionManager.ts
```diff
@@ -73,6 +74,15 @@ export default class ConnectionManager {
...
   }
 
+  public async closeConn() {
+    if (this.conn) {
+      this.getConn().then(conn => {
+        this.isClose = true
+        conn.end()
+      })
+    }
+  }
```
hallettj, line 83: This is good! But you used both the `async` keyword and `.then`. Typically you use one or the other, and I recommend preferring `async` / `await`. So you could change the body of this method to `const conn = await this.getConn(); conn.end()`. (I know that I used `.then` in `getConn`; that is because I wanted to add a callback to the promise, and also hold a reference to the resulting promise.)

You should also move `this.isClose = true` to the top so that it is set immediately and unconditionally.

packages/main/src/resolvers/account.test.ts
```diff
@@ -374,6 +375,25 @@ describe("sync", () => {
       data: { accounts: { delete: false } }
     })
   })
+
+  it("closes a connection to an account upon deleting that account", async () => {
+    // const conn = AccountManager.getConnectionManager(accountId);
```
hallettj: Please remove commented code

packages/main/src/resolvers/account.test.ts
```diff
@@ -374,6 +375,25 @@ describe("sync", () => {
...
+          accounts {
+            delete(id: $id)
+          }
+        }
+      `,
+      null,
+      null,
+      { id: accountId }
+    )
+    expect(AccountManager.isLoggedIn(accountId)).toBe(false)
```
hallettj, line 393: Good stuff

hallettj (changes requested):

hallettj (approved):

PR #8, opened 2019-06-20
Test suite for client

hallettj:
This change sets up Jest and Enzyme for testing React components. It includes fixtures and helpers to mock GraphQL requests.


packages/client/src/App.test.tsx
```diff
@@ -1,9 +1,75 @@
...
 
-it("renders without crashing", () => {
-  const div = document.createElement("div")
-  ReactDOM.render(<App />, div)
-  ReactDOM.unmountComponentAtNode(div)
+it("displays a list of conversations", async () => {
+  const app = mount(<App />, {
+    mocks: [$.getAccountMock],
+    route: "accounts/1/dashboard"
+  })
```
hallettj, line 14: The `mount` test helper renders a React component for testing. It sets up a provider to handle GraphQL queries originating from the component. It also sets up a `LocationProvider` to provide the initial URL set up by the test, and to keep track of any URL changes that components make.

In most cases you will need to provide an array of mocks. Each of these is a predefined response to a GraphQL query. You will need to provide one mock for each query. If the same query runs more than once, you will need to specify a mock for each request.

packages/client/src/App.test.tsx
```diff
@@ -1,9 +1,75 @@
...
+  const row = findConversationRow(app, $.conversation)
+  expect(row).toIncludeText("Hello from test")
+  row.find(Avatar).simulate("click")
+  app.find('button[aria-label="archive"]').simulate("click")
+  await updates(app, 10)
+  expect(findConversationRow(app, $.conversation)).not.toExist()
+})
+
+it("archives a conversation from the conversation view", async () => {
+  const history = createHistory(
```
hallettj, line 41: Poodle uses @reach/router to render different React components when the URL changes. By default the router reads the location from `window.location`. But we can provide a different history source for the router to read and update location. In tests we use a "memory source" which stores the location in a plain javascript object. In this test I create a `history` object explicitly and pass it to `mount` so that I can check `history.location` after I expect the app to have changed the URL.

packages/client/src/App.test.tsx
```diff
@@ -1,9 +1,75 @@
...
+      {
+        ...$.getAccountMock,
+        result: { data: { account: { ...$.account, conversations: [] } } }
+      }
+    ],
+    history
+  })
+  await updates(app)
+  app.find("button[aria-label='archive']").simulate("click")
+  await updates(app, 10)
```
hallettj, line 60: Use the `updates` helper when you need to allow for a delay for the component to update after an asynchronous process, such as a GraphQL request. You can provide a second argument to specify a longer delay, given in milliseconds.

packages/client/src/Conversation.tsx
```diff
@@ -79,14 +84,7 @@ export default function Conversation({ accountId, conversationId }: Props) {
...
   })
-  const [setIsRead, setIsReadResult] = graphql.useSetIsReadMutation({
-    variables: { conversationId: conversationId!, isRead: true }
-  })
-  React.useEffect(() => {
-    if (data && data.conversation && !data.conversation.isRead) {
-      setIsRead()
-    }
-  }, [data, setIsRead])
+  const setIsReadResult = useSetIsRead(data && data.conversation)
```
hallettj, line 87: I decided to simplify the code here by replacing the graphql mutation and effect hook with a single invocation of a custom hook.

packages/client/src/Dashboard.tsx
```diff
@@ -103,7 +103,7 @@ const useStyles = makeStyles(theme => ({
   }
 }))
 
-export default function Dashboard({ accountId }: Props) {
+export default function Dashboard({ accountId, navigate }: Props) {
```
hallettj, line 106: I switched from using the imported `navigate` function to getting `navigate` from props. `Dashboard` is a child component of a `Router` component. `Router` injects the `navigate` prop into its children. The `navigate` function in props uses the mock history set up in `LocationProvider`, which is important for testing.

packages/client/src/Dashboard.tsx
```diff
@@ -239,7 +240,7 @@ function SelectedActionsBar({
       >
         <Toolbar className={classes.toolbar}>
           <span className={classes.title} />
-          <IconButton onClick={onArchive}>
+          <IconButton aria-label="archive" onClick={onArchive}>
```
hallettj, line 243: Adding an `aria-label` here is useful for accessibility, and it also gives me a selector I can use to find the archive button in tests.

packages/client/src/Dashboard.tsx
```diff
@@ -280,100 +281,114 @@ type Conversation = NonNullable<
...
-                ) : (
-                  <Divider variant="inset" component="li" />
-                )}
-              </React.Fragment>
-            )
-          }
-        )}
+        {conversations.map((conversation, index) => {
+          return (
+            <React.Fragment key={conversation.id}>
```
hallettj, line 299: I changed the `key` prop from `index` to `conversation.id`. React uses the `key` prop to match up elements that should be considered "the same" from one render to the next. When performing updates we want React to treat rows representing the same conversation as the same, and using the conversation ID as a key makes this work.

packages/client/src/Dashboard.tsx
```diff
@@ -280,100 +281,114 @@ type Conversation = NonNullable<
...
+              )}
+            </React.Fragment>
+          )
+        })}
       </List>
     </Paper>
   )
 }
 
+function ConversationRow({
```
hallettj, line 314: I split out a new component to display each row in the conversation list view. This way I can easily locate conversation rows in tests with `app.find("ConversationRow")`.

packages/client/src/hooks/useSetIsRead.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from "react"
+import { MutationResult } from "react-apollo"
+import * as graphql from "../generated/graphql"
+
+export default function useSetisRead(
```
hallettj, line 5: This is a custom hook; this pattern is highly useful for encapsulating logic to keep components as simple as possible. A custom hook is a function. The name always starts with `use` so that ESLint can correctly make check uses of the hook.

Notice that a custom hook can call other hooks. Hooks only work when they are called either from function components or from custom hooks.

packages/client/src/testing/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+let mounts: enzyme.ReactWrapper[] = []
+
+afterEach(() => {
+  for (const m of mounts) {
+    m.unmount()
+  }
+  mounts = []
+})
+
+export function mount(
```
hallettj, line 21: This is a helper for rendering React components for tests. Most of our components will only work correctly if they are rendered inside of an Apollo provider component that handles making GraphQL queries. In this case `MockedProvider` sends back predefined responses instead of calling a real API. In addition `LocationProvider` tells @reach/router to use the provided mock history object instead of using `window.history`, which would attempt to use the real window location.

There is an `afterEach` callback here that automatically unmounts any mounted components after each test.

breitman (approved):

PR #9, opened 2019-06-24
Br/travis

breitman:


hallettj:
According to what I have read the `no version information available` warning happens because we are running a pre-built binary (probably electron) that is compiled against a newer version of a shared library than what is installed on Travis. Those messages might go away if we configure Travis to use a newer Ubuntu version. The latest version that Travis supports is Ubuntu Xenial (16.04) (see [here](https://docs.travis-ci.com/user/reference/overview/)). There are instructions for setting the build dist [here](https://docs.travis-ci.com/user/reference/xenial/).

Edit: using Ubuntu 16.04 might also fix the `extension "RANDR" missing` message according to [this post](https://askubuntu.com/a/773379/14946).

.travis.yml
```diff
@@ -0,0 +1,12 @@
...
+  - "12"
+before_install:
+  - "sudo apt-get install -y libgnome-keyring-dev libsecret-1-dev"
+  - "sudo apt-get install -y xvfb"
+
+script: xvfb-run yarn test
+
+  # before_script:
+  # - "export DISPLAY=:99.0"
+  # - "sh -e /etc/init.d/xvfb start"
```
hallettj: It looks like tests are running, and this `before_script` block is not necessary.

scripts/tests
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+(cd packages/main && eslint src/**/*.ts)
+(cd packages/client && eslint src/**/*.ts)
```
hallettj: It turns out that the eslint command I gave you misses problems in `.tsx` files. You can use these commands instead:

```sh
(cd packages/main && eslint src/**/*.ts src/**/*.tsx)
(cd packages/client && eslint src/**/*.ts src/**/*.tsx)
```

hallettj (changes requested):

hallettj:
In [a previous iteration of poodle](https://github.com/PoodleApp/poodle_iteration-2/blob/master/.travis.yml) I used this Travis configuration to cache package builds:

```yaml
env:
  global:
    - ELECTRON_CACHE=$HOME/.cache/electron
    - ELECTRON_BUILDER_CACHE=$HOME/.cache/electron-builder

cache:
  yarn: true
  directories:
    - $HOME/.cache/electron
    - $HOME/.cache/electron-builder
    - $HOME/.electron
    - $HOME/.electron-gyp
    - $HOME/.node-gyp
    - $HOME/.npm/_prebuilds
```

We can run tests in a MacOS environment in addition to a Linux environment with this config:

```yaml
os:
  - linux
  - osx
```

That might be helpful to prevent me from accidentally breaking your development environment. If you make that change you will need to change the way that apt packages are installed so that the `apt-get install` step only runs in the Linux environment. You can do that by replacing the `before_install` block with an invocation of the `apt` addon like this:

```yaml
addons:
  apt:
    packages:
      - libgnome-keyring-dev
      - libsecret-1-dev
      - xvfb
```

PR #14, opened 2019-07-01
Handle cases where email addresses from Google Contacts can't be parsed

### LINKED ISSUE #12: Some contacts from Google do not provide valid email addresses
We are getting an error saving Google contacts where the `build` function from `models/Address` returns an address with an `undefined` host. This is because Google is sending records with a `value` property that is not actually an email address. We should skip over such contacts, and not attempt to store them in the database.

hallettj:
resolves #12

### DIFF
--- a/packages/main/src/cache/persist.ts
+++ b/packages/main/src/cache/persist.ts
@@ -3,8 +3,6 @@ import db from "../db"
 import { MessageAttributes } from "../types"
 import { getPartByPartId } from "./query"
 import { ID, SerializedHeaders } from "./types"
-import { people_v1 } from "googleapis"
-import { build } from "../models/Address"
 
 export function persistBoxState(accountId: ID, box: imap.Box): ID {
   return db.transaction(() => {
@@ -253,28 +251,13 @@ export function persistTokens(syncToken: string, accountId: ID) {
   )
 }
 
-export function persistContacts(
+export function persistContact(
   accountId: ID,
-  connections: people_v1.Schema$Person[]
+  contact: { name?: string; mailbox: string; host: string }
 ) {
-  connections.forEach(connection => {
-    connection.emailAddresses &&
-      connection.emailAddresses.forEach(email => {
-        const contact = build({
-          email: email.value!,
-          name:
-            connection.names &&
-            connection.names[0] &&
-            connection.names[0].displayName
-        })
-
-        insertInto("google_connections", {
-          account_id: accountId,
-          host: contact.host,
-          mailbox: contact.mailbox,
-          name: contact.name
-        })
-      })
+  insertInto("google_connections", {
+    account_id: accountId,
+    ...contact
   })
 }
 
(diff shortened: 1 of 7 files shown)


packages/main/src/cache/persist.ts
```diff
@@ -253,28 +251,13 @@ export function persistTokens(syncToken: string, accountId: ID) {
...
 
-export function persistContacts(
+export function persistContact(
   accountId: ID,
-  connections: people_v1.Schema$Person[]
+  contact: { name?: string; mailbox: string; host: string }
 ) {
-  connections.forEach(connection => {
-    connection.emailAddresses &&
-      connection.emailAddresses.forEach(email => {
```
hallettj, line 262: I would rather keep the `cache/` modules focused on moving data into and out of the database, and keep parsing and conversion logic in other places.

packages/main/src/contacts.test.ts
```diff
@@ -1,18 +1,12 @@
-import db from "./db"
```
hallettj, line 1: There were a bunch of unused imports here.

packages/main/src/contacts.ts
```diff
@@ -36,8 +35,19 @@ export default class ContactsApiClient {
 
   async downloadContacts(accountId: ID) {
     const connections = await this.getContacts(accountId)
-    if (connections) {
-      persistContacts(accountId, connections)
+    for (const connection of connections || []) {
+      for (const email of connection.emailAddresses || []) {
```
hallettj, line 39: I think that `for ... of` is more idiomatic than `.forEach()`. It used to be that there was no great way to iterate over arrays, and `.forEach()` was the way to go. But since `for ... of` was added in ES6 I prefer to use it. An important advantage is that while `.forEach()` only works on arrays (or other values that implement a method with the same name) `for ... of` can iterate over any iterable value: arrays, argument lists, DOM node lists, Immutable.js collections, etc.

packages/main/src/models/Address.ts
```diff
@@ -44,15 +44,17 @@ export function build({
...
-  const [mailbox, host] = email.split("@", 2)
-  return name
-    ? new Address({
-        name,
-        mailbox,
-        host
-      })
-    : new Address({ mailbox, host })
+}): Address | null {
+  const parsed = addrs.parseOneAddress(email)
```
hallettj, line 48: My solution to the bug of invalid email addresses is to use a proper email parsing library instead of the quick solution I had previously. `parseOneAddress` will return `null` if its input is not an email address.

packages/main/src/queue/combineHandlers.ts
```diff
@@ -125,11 +125,13 @@ function onFailure<HM extends Handlers>(
   error: JobFailure<HM>,
   _stats: { elapsed: number }
 ) => Promise<void> {
-  return async (_taskId, { cause, task }) => {
-    const failureHandler = handlers[task.type].failure
-    if (failureHandler) {
-      await failureHandler(cause, task.params)
-    }
+  return async (_taskId, error) => {
+    console.error(error)
```
hallettj: I made a change here because this code did not work as I thought it did, and I wanted to have a descriptive message in the console when a queued task fails.

breitman (approved):

PR #17, opened 2019-07-02
Br/reply to

breitman:
In test file, the 'from' field says name is null. Not sure why that is, but I matched it in the expect block to let the test pass. It must be due to the beforeEach block that adds the same account without including a name field. I will change it if needed, but I am pretty sure this fix is working correctly.


packages/main/src/compose/edit.ts
```diff
@@ -37,19 +37,22 @@ export function composeEdit({
...
       date: date,
       envelope: {
         date,
         subject: subject ? `Re: ${subject}` : null,
-        from: from.toArray(),
-        sender: from.toArray(),
-        replyTo: from.toArray(),
-        to: to.toArray(),
-        cc: cc.toArray(),
+        from: from!,
```
hallettj: It's not a good idea to use a non-null assertion (`!`) when you know that the value might be `null` or `undefined`. You might have put the assertion in place because of a message like `undefined is not assignable to Address[] | null`. Instead of using the assertion, please produce a value that is assignable to the expected type.

packages/main/src/compose/edit.ts
```diff
@@ -37,19 +37,22 @@ export function composeEdit({
...
+        sender: from!,
+        replyTo: replyTo ? replyTo : null,
```
hallettj: This is a good way to handle the issue that `replyTo` might be `undefined`. It can be made slightly shorter: `replyTo || null`.

packages/main/src/compose/reply.ts
```diff
@@ -18,19 +18,23 @@ export function composeReply({
...
       date: date,
       envelope: {
         date,
         subject: subject ? `Re: ${subject}` : null,
-        from: from.toArray(),
-        sender: from.toArray(),
-        replyTo: from.toArray(),
-        to: to.toArray(),
-        cc: cc.toArray(),
+        from: from!,
```
hallettj: Same deal: please don't use non-null assertions in this case.

packages/main/src/models/Address.ts
```diff
@@ -6,7 +6,7 @@ import { URI, mailtoUri } from "./uri"
 export type Email = string
 
 export default class Address {
-  name: string | undefined
+  name: string | null | undefined
```
hallettj, line 9: Good thinking

packages/main/src/models/conversation.ts
```diff
@@ -13,11 +13,12 @@ export interface Conversation {
...
-  from: Collection.Indexed<imap.Address>
-  to: Collection.Indexed<imap.Address>
-  cc: Collection.Indexed<imap.Address>
-}
+// type Participants = {
+//   from?: Collection.Indexed<imap.Address>
+//   to: Collection.Indexed<imap.Address>
+//   cc: Collection.Indexed<imap.Address>
+//   replyTo: Collection.Indexed<imap.Address>
+// }
```
hallettj: Please delete commented code.

packages/main/src/models/conversation.ts
```diff
@@ -41,29 +42,56 @@ export function getReplyParticipants(
...
+      )
     )
-  ).sortBy(Addr.formatAddress)
-  return { to, cc, from: Seq([senderAddress]) }
+  )
+    .sortBy(Addr.formatAddress)
+    .toArray()
+
+  // console.log(participants.replyTo)
+  // const replyTo = participants.replyTo
```
hallettj: Please remove commented code

packages/main/src/models/conversation.ts
```diff
@@ -41,29 +42,56 @@ export function getReplyParticipants(
...
+
   const to = uniqBy(
     Addr.normalizedEmail,
-    participants.from
-      .concat(participants.to)
-      .filter(p => !Addr.equals(senderAddress, p))
-  ).sortBy(Addr.formatAddress)
+    List(
+      participants
+        .replyTo!.concat(participants.to)
```
hallettj: This is not an acceptable use of a non-null assertion. The type checker is trying to help you here - don't fight it! As it is written now I understand that `participants.replyTo` will always be defined. But that could easily change in the future. Type assertions eliminate opportunities for the type checker to catch future bugs.

packages/main/src/models/conversation.ts
```diff
@@ -41,29 +42,56 @@ export function getReplyParticipants(
...
+  )
+    .sortBy(Addr.formatAddress)
+    .toArray()
+
+  // console.log(participants.replyTo)
+  // const replyTo = participants.replyTo
+  const replyTo = uniqBy(
+    Addr.normalizedEmail,
+    List(participants.replyTo!.filter(p => !Addr.equals(senderAddress, p)))
+  ).toArray()
```
hallettj: Keep in mind the goal: we are building headers for a reply. It is sent from the address of the current account. The `replyTo` value should be an address that someone can use later to send a message to the author of this reply.

Also remember that a `replyTo` header is not required if its value is the same as the `from` value.

packages/main/src/models/conversation.ts
```diff
@@ -41,29 +42,56 @@ export function getReplyParticipants(
...
+    Seq(conversation.messages)
+      .flatMap(message => {
+        if (type === "replyTo") {
+          const replyParts = cache.getParticipants(message.id, type)
+          return replyParts.length !== 0
+            ? replyParts
+            : cache.getParticipants(message.id, "from")
+        }
+        return cache.getParticipants(message.id, type)
+      })
```
hallettj, line 75: Looks good!

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -502,114 +611,138 @@ it("sends an edit", async () => {
...
-    editedMessage: { envelope_messageId: message.envelope.messageId },
-    editedPart: {
-      content_id: part.id
-    },
-    resource: {
-      messageId: message.envelope.messageId,
-      contentId: part.id
-    }
+describe("no before sync test", () => {
+  beforeEach(async () => {
```
hallettj: Notice that all of the code in this `beforeEach` call is also present in the `beforeEach` call in the other `describe` block. This `beforeEach` should be moved to the top of the file. A `beforeEach` call outside of any `describe` block will run before any `beforeEach` calls in `describe` blocks.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -22,477 +22,586 @@ let connectionManager: ConnectionManager
 let conversation: Conversation
 let conversationId: string
```
hallettj: You moved the assignment for some of these variables to a `beforeEach` inside a `describe` block. Those variables will not be assigned properly while the other `describe` block runs, and so should not be referenced from the other block. To enforce that please move the variable declarations to the `describe` blocks where they are assigned.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -502,114 +611,138 @@ it("sends an edit", async () => {
...
-            editedBy {
-              name
-              mailbox
-              host
+  it("uses the replyTo field instead of from when replyTo is provided", async () => {
+    mock(Connection.prototype.fetch).mockImplementation(
+      mockFetchImplementation({
+        thread: [
+          testThread[0],
+          // testThread[1],
```
hallettj: Please remove commented code

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -502,114 +611,138 @@ it("sends an edit", async () => {
...
-              name
-              mailbox
-              host
-            }
-            editedAt
-            editedBy {
-              name
-              mailbox
-              host
+  it("uses the replyTo field instead of from when replyTo is provided", async () => {
```
hallettj: This test title could use more context. What uses the replyTo field? For what purpose does it use the replyTo field? Test titles do need to be short, but they also need to inform the user generally what area of responsibility the test applies to. This could be "addresses replies using replyTo header when available"; or you could provide context using a describe block like this:

```ts
describe("when addressing replies", () => {
  it("sends to a previous message's replyTo address when one is provided", async () => {
```

Describe blocks are quite useful for providing context in large test suites. When writing titles a test title states what something does (`it("does a thing")`), and a describe title says when or under what conditions those statements apply (`describe("when doing such-and-such a thing")`). And describe blocks can be nested in other describe blocks.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -502,114 +611,138 @@ it("sends an edit", async () => {
...
+          {
+            ...testThread[1],
+            attributes: {
+              ...testThread[1].attributes,
+              envelope: {
+                ...testThread[1].attributes.envelope,
+                replyTo: [
+                  { name: "Jesse Hallett", mailbox: "jesse", host: "test.com" }
+                ]
+              }
```
hallettj, line 713: Good stuff!

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -502,114 +611,138 @@ it("sends an edit", async () => {
...
+                host: "gmail.com"
+              },
+              {
+                name: "Jesse Hallett",
+                mailbox: "jesse",
+                host: "test.com"
+              }
+            ],
+            cc: [],
+            replyTo: [
```
hallettj: This is not the value we want for `replyTo`. When someone wants to reply to *this* message, what address should the send to? hallettj@gmail.com is the `from` / `replyTo` address for some other message, not for this one.

packages/main/src/resolvers/conversation.ts
```diff
@@ -47,7 +47,7 @@ export const Conversation: ConversationResolvers = {
   replyRecipients(conversation: C.Conversation, { fromAccountId }) {
     const account = mustGetAccount(fromAccountId)
     const recipients = C.getReplyParticipants(conversation, account)
-    return mapObject(recipients, rs => rs.toArray())
+    return recipients
```
hallettj, line 50: Good catch.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -502,114 +611,138 @@ it("sends an edit", async () => {
...
-                subtype: "plain",
-                content: revisedContent
+                name: "Jesse Hallett",
+                mailbox: "hallettj",
+                host: "gmail.com"
+              },
+              {
+                name: "Jesse Hallett",
+                mailbox: "jesse",
+                host: "test.com"
```
hallettj, line 769: It's not ideal to send a message back to our own replyTo address. But we don't have logic yet for actually sending messages with a replyTo header, so this is good for now.

hallettj (changes requested):
I appreciate all of the effort you are putting into this feature! I know that the work has been complicated; I'm pleased with your persistence.

Yes, the `name` for some addresses will be `null`. That is because I have not added support for setting a name for an account, so messages sent from Poodle do not have names in the `from` address.

In the future I would like you to make pull request titles more descriptive. It's a good habit because descriptive titles help code reviewers to understand what they are looking at, and pull requests form a historical record.

There are merge conflicts - you will need to merge changes from master and resolve those conflicts.

hallettj:
I'm seeing a number of unused variable warnings from eslint. Please make a habit of using the "organize imports" feature in VS Code to remove unused imports, and to put import lines into a consistent order. And if you have not already done so it would be helpful to configure your editor to display eslint warnings.

packages/main/src/resolvers/conversation.test.ts
hallettj: This is still not correct. Who is this message from? Why is the replyTo value different from the from value?

hallettj (changes requested):
There are some remaining things that I would like you to change. And it looks like the merge introduced a problem. When you are ready we can talk more about how replyTo should be produced.

packages/main/src/models/conversation.ts
```diff
@@ -35,28 +35,33 @@ export function getReplyParticipants(
...
-    Addr.normalizedEmail,
-    List(
-      participants
-        .replyTo!.concat(participants.to)
-        .filter(p => !Addr.equals(senderAddress, p))
-    )
-  )
-    .sortBy(Addr.formatAddress)
-    .toArray()
+  const to = participants.replyTo
```
hallettj: This is good thinking, but what if `participants.to` is defined? In that case we want to use those values for the `to` list for the new message.

I would move the ternaries to more specific sites like this:

```ts
uniqBy(
        Addr.normalizedEmail,
        List(
          (participants.replyTo || [])
            .concat(participants.to || [])
            .filter(p => !Addr.equals(senderAddress, p))
        )
      )
```

packages/main/src/models/conversation.ts
```diff
@@ -35,28 +35,33 @@ export function getReplyParticipants(
...
+        .toArray()
+    : []
 
-  const cc = uniqBy(
-    Addr.normalizedEmail,
-    List(
-      participants.cc.filter(
-        p =>
-          !Addr.equals(senderAddress, p) && !to.some(p_ => Addr.equals(p, p_))
+  const cc = to
```
hallettj: You know that `to` will be defined, even if it might be an empty array, so this ternary is not necessary. Remember that `[]` is truthy, so the false branch of this conditional will never run.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -18,18 +18,23 @@ jest.mock("imap")
...
 
-describe("same sync tests", () => {
+beforeEach(async () => {
+  const { lastInsertRowid } = db
+    .prepare("insert into accounts (email) values (?)")
+    .run("jesse@sitr.us")
+  accountId = lastInsertRowid
+
+  connectionManager = mockConnection()
+})
```
hallettj, line 29: Good stuff

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -18,18 +18,23 @@ jest.mock("imap")
...
+describe("when addressing conversations", () => {
+  let conversation: Conversation
+  let conversationId: string
+
   beforeEach(async () => {
-    const { lastInsertRowid } = db
-      .prepare("insert into accounts (email) values (?)")
-      .run("jesse@sitr.us")
-    accountId = lastInsertRowid
     account = { id: accountId, email: "jesse@sitr.us" }
```
hallettj: Please either move this assignment to the `beforeEach` above, or move the `account` variable declaration into this `describe` block.

hallettj (changes requested):

hallettj (approved):

PR #19, opened 2019-07-03
sync mail and contacts in parallel 

### LINKED ISSUE #16: sync mail and contacts in parallel
Start syncing mail and contacts without `await`ing. Then use `Promise.all()` to await results of all promises.

breitman:
resolves #16

### DIFF
--- a/packages/main/src/sync.ts
+++ b/packages/main/src/sync.ts
@@ -48,7 +48,7 @@ export async function sync(accountId: cache.ID, manager: ConnectionManager) {
   const contactSync =
     contactApiClient && contactApiClient.downloadContacts(accountId)
 
-  Promise.all([boxSyncer(accountId, manager), contactSync])
+  await Promise.all([boxSyncer(accountId, manager), contactSync])
 }
 
 class BoxSync {
--- a/packages/main/src/sync.ts
+++ b/packages/main/src/sync.ts
@@ -9,6 +9,7 @@ import { publishMessageUpdates } from "./pubsub"
 import * as request from "./request"
 import * as kefirUtil from "./util/kefir"
 import AccountManager from "./managers/AccountManager"
+import { promised } from "q"
 
 type R<T> = kefir.Observable<T, Error>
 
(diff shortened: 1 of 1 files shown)


packages/main/src/sync.ts
```diff
@@ -9,6 +9,7 @@ import { publishMessageUpdates } from "./pubsub"
 import * as request from "./request"
 import * as kefirUtil from "./util/kefir"
 import AccountManager from "./managers/AccountManager"
+import { promised } from "q"
```
hallettj: I think you did not end up using `promised`? Can you remove this line?

packages/main/src/sync.ts
```diff
@@ -25,23 +26,33 @@ const cachePolicy = {
 
 const BATCH_SIZE = 50
 
-export async function sync(accountId: cache.ID, manager: ConnectionManager) {
+async function boxSyncer(accountId: cache.ID, manager: ConnectionManager) {
+  let boxSync
+  let box
```
hallettj: It looks like you started something involving assigning promises to variables and ended up going in a different direction. These variables do not need to be declared outside of the loop.

Because `boxSyncer` is marked `async` it will return a promise regardless of whether there is a return statement. The promise that is returned will not resolve until the promises from all `await` expressions in the function resolve.

packages/main/src/sync.ts
```diff
@@ -25,23 +26,33 @@ const cachePolicy = {
...
+
+export async function sync(accountId: cache.ID, manager: ConnectionManager) {
   const contactApiClient = AccountManager.getContactsApiClient(
     String(accountId)
   )
-  contactApiClient && (await contactApiClient.downloadContacts(accountId))
+  const contactSync =
+    contactApiClient && (await contactApiClient.downloadContacts(accountId))
+
+  Promise.all([boxSyncer(accountId, manager), contactSync])
```
hallettj: This is good, but it is not parallel yet because you `await`ed on the `downloadContacts` call.

hallettj (changes requested):

hallettj (changes requested):
Looking good! There is one more thing you need to do: you want to suspend execution until the parallel operations are complete. `Promise.all` returns a promise that resolves when its inputs resolve. But it will not suspend execution unless you `await`.

hallettj (approved):

PR #20, opened 2019-07-08
conversation has collapsed read messages

breitman:


packages/client/src/Conversation.tsx
```diff
@@ -66,6 +69,16 @@ const useStyles = makeStyles(theme => ({
   edited: {
     fontStyle: "italic"
   },
+  expand: {
+    transform: "rotate(0deg)",
+    marginLeft: "auto",
+    transition: theme.transitions.create("transform", {
+      duration: theme.transitions.duration.shortest
+    })
```
hallettj, line 78: This is nifty

packages/client/src/Conversation.tsx
```diff
@@ -186,6 +199,8 @@ function Presentable({
 }) {
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
   const [editing, setEditing] = React.useState(false)
+  const [expanded, setExpanded] = React.useState(!presentable.isRead || false)
```
hallettj: The `|| false` does not add anything - `!presentable.isRead` is a boolean expression so it evaluates to either `true` or `false`, and either way `|| false` does not produce a different value.

packages/client/src/Conversation.tsx
```diff
@@ -196,6 +211,10 @@ function Presentable({
     setAnchorEl(null)
   }
 
+  function handleExpandClick() {
+    setExpanded(!expanded)
```
hallettj: A one-line callback like this could be in-lined

packages/client/src/Conversation.tsx
```diff
@@ -216,6 +235,17 @@ function Presentable({
             >
               <MoreVertIcon />
             </IconButton>
+            <IconButton
+              className={clsx(classes.expand, {
+                [classes.expandOpen]: expanded
+              })}
+              onClick={handleExpandClick}
+              aria-expanded={expanded}
```
hallettj, line 247: It is helpful to have these aria attributes. To show a user agent *what* has been expanded the proper thing to do is to also include an `aria-controls` attribute set to the `id` of the card content element. You can find more information here: https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions

packages/client/src/Conversation.tsx
```diff
@@ -235,24 +265,25 @@ function Presentable({
...
-                accountId={accountId}
-                conversationId={conversationId}
-                contentToEdit={content}
-                onComplete={() => {
-                  setEditing(false)
-                }}
-              />
-            )
-          } else {
+      <Collapse in={expanded} timeout="auto" unmountOnExit>
```
hallettj: `unmountOnExit` removes child elements from the DOM after the transition ends, as opposed to hiding them. One implication is that when the element is shown again its state is reset, and any queries it runs on mount need to re-run. In the case of presentables if you start editing a presentable, change the text, collapse, and then expand your changes will be lost. So I think we would be better of removing `unmountOnExit`.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -638,6 +638,116 @@ describe("when addressing conversations", () => {
     })
   })
 
+  it("an edited message will have an isRead field", async () => {
```
hallettj: Think of `it` as the first word in the sentence that is the test description. And what does it mean that an edited message will have an isRead field? All presentable elements have an isRead field, regardless of whether they have been edited.

What I think you want to specify in this test is how `isRead` is computed in the case of an edited message. Is the presentable considered to be read if the latest edit is read? Or if the original revision is read? Or only if all revisions have been read? In this test it looks like both the original revision and the edited revision are unread, so the test does not answer that question.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -638,6 +638,116 @@ describe("when addressing conversations", () => {
...
+                subtype
+                content
+              }
+            }
+          }
+        }
+      `,
+      { conversationId: testThread[0].attributes["x-gm-thrid"] }
+    )
+    expect(result).toMatchObject({
```
hallettj, line 688: This test could be made much shorter by requesting fewer fields, and only specifying the one field that you care about in the expect matcher.

packages/client/src/Conversation.test.tsx
```diff
@@ -29,3 +30,18 @@ it("displays a loading indicator while the conversation is leading", () => {
   )
   expect(app).toIncludeText("Loading...")
 })
+
+it("collapses read messages in a conversation", () => {
```
hallettj: In its present form this test does not make any assertions about whether a card is collapsed. A good practice to follow is the "red-green" cycle: write a test first, make sure that it fails, write code until the test passes. Writing the test first is ideal; but the next best thing is to temporarily disable the feature you worked on and make sure that the test fails. I noticed that this test passes even if `expanded` is always set to `true` or to `false`.

What would be great is a test that explicitly checks that a read presentable is collapsed, and that an unread presentable is not collapsed.

packages/client/src/Conversation.test.tsx
```diff
@@ -29,3 +30,18 @@ it("displays a loading indicator while the conversation is leading", () => {
...
+  const app = mount(
+    <Conversation
+      accountId={$.account.id}
+      conversationId={$.conversation.id}
+    />,
+    {
+      mocks: [$.getConversationMock, $.setIsReadMock]
+    }
+  )
+  updates(app)
```
hallettj: I think you forgot to put an `await` in this `updates` call.

hallettj (changes requested):
This feature is looking quite nice! But the tests need some more work.

packages/client/src/Conversation.test.tsx
```diff
@@ -41,7 +41,37 @@ it("collapses read messages in a conversation", () => {
       mocks: [$.getConversationMock, $.setIsReadMock]
     }
   )
-  updates(app)
+  await updates(app)
 
-  expect(app.find(CardContent)).toBeTruthy()
+  expect(app.find(Collapse)).toHaveProp("in", true)
```
hallettj: This is a good assertion!

packages/client/src/Conversation.test.tsx
```diff
@@ -41,7 +41,37 @@ it("collapses read messages in a conversation", () => {
...
+
+  const newConvo = {
```
hallettj: Since you are setting up a new set of test data and a new instance of `Conversation` I think this should go into a new test.

packages/client/src/Conversation.tsx
```diff
@@ -239,9 +235,10 @@ function Presentable({
               className={clsx(classes.expand, {
                 [classes.expandOpen]: expanded
               })}
-              onClick={handleExpandClick}
+              onClick={() => setExpanded(!expanded)}
               aria-expanded={expanded}
               aria-label="Show more"
+              aria-controls="message-card-content"
```
hallettj: This is good - but the id needs to be unique. There are likely to be multiple presentable elements on a page.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -638,7 +638,7 @@ describe("when addressing conversations", () => {
     })
   })
 
-  it("an edited message will have an isRead field", async () => {
+  it("marks an edited message as read", async () => {
```
hallettj: This title and the test itself are much improved! But I take issue with the word "marks" - that implies a state change, but what you are testing is an interpretation of some given data. I would say "interprets an edited message..." or "treats an edited message as...". But that still does not accurately describe what you are testing. It is possible for an edited message to be unread. This test specifies that the read status of a presentable is based on the read status of the most recent revision.

hallettj (changes requested):

packages/client/src/Conversation.tsx
```diff
@@ -244,7 +244,7 @@ function Presentable({
               onClick={() => setExpanded(!expanded)}
               aria-expanded={expanded}
               aria-label="Show more"
-              aria-controls="message-card-content"
+              aria-controls={`${presentable.id}-card-content`}
```
hallettj: Yes, this is good. In the future I recommend putting the id at the end of the string because it is very common in programming languages and markup languages to have a requirement where identifiers must begin with a letter. An HTML id is a form of identifier, and `presentable.id` happens to be a string made up of numeric characters. In fact HTML 4 did not allow ids to begin with numbers - but HTML 5 does, so this usage is ok. See https://stackoverflow.com/a/79022/103017

I only bring this up because the format of this id activated my developer spider sense because the rule that identifiers must not begin with numbers is very common.

packages/client/src/Conversation.tsx
```diff
@@ -269,7 +269,7 @@ function Presentable({
         }
       />
       <Collapse in={expanded} timeout="auto">
-        <CardContent id="message-card-content">
+        <CardContent id={`${presentable.id}-card-content`}>
```
hallettj: In cases like this where it is important that two strings are identical it is useful to define a variable inside of writing the string twice. A variable can make your intention more clear to developers who come along later who might otherwise change one string and not know that the other string should be changed to match. Just something to keep in mind in the future.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -638,7 +638,7 @@ describe("when addressing conversations", () => {
     })
   })
 
-  it("marks an edited message as read", async () => {
+  it("uses the read status of a message's most recent revision as the read status of the presentable", async () => {
```
hallettj, line 641: Good stuff

hallettj (approved):

PR #22, opened 2019-07-09
archived message should be unarchived locally on failure to update server

breitman:


packages/main/src/cache/persist.ts
```diff
@@ -396,6 +396,32 @@ export function delLabels({
...
+}: {
+  accountId: ID
+  box: { name: string }
+  uids: number[]
+  labels: string[]
+}) {
+  labels.forEach(label => {
+    db.prepare(
+      `
+      insert into message_gmail_labels (id, label, message_id)
```
hallettj: When inserting please don't provide a value for `id`. The `id` column is declared `integer primary key` which in sqlite means that the database will automatically assign an `id` when the record is created, and the database can do a better job than application code of choosing an `id` value that is not already in use.

The `select` clause here selects three values: `@accountId`, `@label`, `messages.id`, and inserts those values into the columns `id`, `label`, `message_id` in the new rows. The label and message ID assignments are correct, but we do not want to use the account ID as the `id` of the new label row.

Otherwise this query is looking good.

packages/main/src/queue/index.test.ts
```diff
@@ -229,6 +229,43 @@ it("removes cached message if message cannot be sent", async () => {
   })
 })
 
+it("removes archive state changes from cache on failure", async () => {
```
hallettj, line 232: Nice test!

packages/main/src/queue/index.ts
```diff
@@ -80,6 +80,20 @@ const handlers = {
...
+        accountId,
+        box,
+        uids
+      }: {
+        accountId: ID
+        box: { name: string }
+        uids: number[]
+      }
+    ) {
+      cache.addLabels({ accountId, box, uids, labels: ["\\Inbox"] })
```
hallettj, line 96: perfect!

hallettj (changes requested):

packages/main/src/cache/persist.ts
```diff
@@ -410,7 +410,7 @@ export function addLabels({
   labels.forEach(label => {
     db.prepare(
       `
-      insert into message_gmail_labels (id, label, message_id)
+      insert into message_gmail_labels (label, message_id)
```
hallettj, line 413: You'll notice that you got a test failure after this change:

> UnhandledPromiseRejectionWarning: SqliteError: 3 values for 2 columns

Every value that you `select` is used as an inserted value. You declare that you are going to insert values for `label` and `message_id`, and you select `@accountId`, `@label`, `messages.id`. Sqlite attempts to insert the value `@accountId` for the `label` of the new row, the value `@label` for `message_id`, and then notices that you did not provide a slot for the value `messages.id` to be inserted.

hallettj (changes requested):

hallettj (approved):
That's much better - thanks!

PR #25, opened 2019-07-11
remove read and unread state changes locally from cache on failure to update server

breitman:


hallettj (changes requested):
This is great! Would you also add a failure callback for the `markAsRead` handler, and add a corresponding test?

hallettj (approved):
Great, thanks!

PR #26, opened 2019-07-12
a composed message has a content ID

breitman:


packages/main/src/compose/edit.test.ts
```diff
@@ -89,6 +89,64 @@ it("edits a message", () => {
...
+    editedMessage,
+    editedPart,
+    resource: {
+      messageId: testMessage.envelope.messageId,
+      contentId: testPart.id
+    }
+  })
+
+  expect(attributes).toMatchObject({
+    struct: [
```
hallettj, line 107: I would rather remove some properties from this expected object. I think that just testing for the `id`, `type`, and `subtype` properties would be best. The `id` is the property that is being tested; `type` and `subtype` provide just enough context.

packages/main/src/compose/edit.test.ts
```diff
@@ -97,3 +155,7 @@ function conversationFrom(
   }
   return cache.getThreads(accountId)[0]
 }
+
+afterEach(async () => {
+  db.prepare("delete from accounts").run()
```
hallettj, line 140: Good catch

packages/main/src/compose/newConversation.test.ts
```diff
@@ -0,0 +1,45 @@
...
+let accountId: cache.ID
+let boxId: cache.ID
+
+beforeEach(async () => {
+  const { lastInsertRowid } = db
+    .prepare("insert into accounts (email) values (?)")
+    .run("jesse@sitr.us")
+  accountId = lastInsertRowid
+  account = cache.getAccount(accountId)!
+  boxId = cache.persistBoxState(accountId, inbox)
```
hallettj: The function for composing a new conversation does not depend on cache state, so there is no need to call `persistBoxState` or to assign a `boxId` variable.

Actually this test file does not need any database setup at all. You can make up a value for `account` that is compatible with the cache record shape like this:

```ts
const account = { email: "jesse@sitr.us", id: 1 }
```

packages/main/src/compose/newConversation.test.ts
```diff
@@ -0,0 +1,45 @@
...
+})
+describe("a message for a new Conversation", () => {
+  it("has a content-ID", async () => {
```
hallettj, line 5: Good stuff! But as with the other tests we do not need to specify `partID` or `params` in the expected object.

packages/main/src/compose/reply.test.ts
```diff
@@ -85,6 +85,34 @@ describe("reply recipients", () => {
   })
 })
 
+describe("reply", () => {
+  it("has a Content-ID header", async () => {
```
hallettj, line 89: nice

hallettj (changes requested):

hallettj (approved):
Looks good! I restarted the tests, because I cannot reproduce the test error locally.

hallettj:
Oh, the tests are failing after merging changes in this PR with master. Please merge from master and fix the error before merging.

hallettj (approved):
Great, thanks!

PR #27, opened 2019-07-12
Open conversation when clicking a conversation link

hallettj:
In a previous PR I added a feature that allows you to insert links in a message to other conversations using an autocomplete popup. This change handles clicks on those links by opening the referenced conversation.

You will need to delete your cache for this to work because I changed the way that message ID values are stored in the database.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "workspaces": [
     "packages/client",
+    "packages/common",
     "packages/main"
   ],
   "scripts": {
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -6,7 +6,9 @@
   "author": "Jesse Hallett <jesse@sitr.us>",
   "license": "Apache-2.0",
   "scripts": {
+    "prestart": "cd ../common && yarn build",
     "start": "cross-env BROWSER=none react-scripts start",
+    "prebuild": "cd ../common && yarn build",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
--- a/packages/client/src/Conversation.test.tsx
+++ b/packages/client/src/Conversation.test.tsx
@@ -43,19 +43,15 @@ it("collapses read messages, with the exception of the last message in the conve
   )
   await updates(app)
 
-  expect(
-    app
-      .find("Presentable")
-      .filterWhere(n => n.props().presentable.id === "11")
-      .find(Collapse)
-  ).toHaveProp("in", false)
+  expect(app.find({ presentable: { id: "11" } }).find(Collapse)).toHaveProp(
+    "in",
+    false
+  )
 
-  expect(
-    app
-      .find("Presentable")
-      .filterWhere(n => n.props().presentable.id === "12")
-      .find(Collapse)
-  ).toHaveProp("in", true)
+  expect(app.find({ presentable: { id: "12" } }).find(Collapse)).toHaveProp(
+    "in",
+    true
+  )
 })
 
 it("does not collapse unread messages", async () => {
@@ -90,10 +86,8 @@ it("does not collapse unread messages", async () => {
   )
   await updates(app)
 
-  expect(
-    app
-      .find("Presentable")
-      .filterWhere(n => n.props().presentable.id === "11")
-      .find(Collapse)
-  ).toHaveProp("in", true)
+  expect(app.find({ presentable: { id: "11" } }).find(Collapse)).toHaveProp(
+    "in",
+    true
+  )
 })
--- a/packages/client/src/DisplayContent.tsx
+++ b/packages/client/src/DisplayContent.tsx
@@ -1,6 +1,8 @@
 import { makeStyles } from "@material-ui/styles"
+import { Location } from "@reach/router"
 import clsx from "clsx"
 import marked from "marked"
+import { parseMidUri } from "poodle-common/lib/models/uri"
 import * as React from "react"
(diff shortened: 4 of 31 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "workspaces": [
     "packages/client",
+    "packages/common",
```
hallettj, line 6: I created a new package, common, which can be imported in both client and main using `import whatever from "poodle-common/whatever`.

packages/client/package.json
```diff
@@ -6,7 +6,9 @@
   "author": "Jesse Hallett <jesse@sitr.us>",
   "license": "Apache-2.0",
   "scripts": {
+    "prestart": "cd ../common && yarn build",
```
hallettj, line 9: The new common package needs to be compiled so that the other packages can use it. I put in some script hooks here to automatically compile common when running `yarn start` from client.

packages/client/src/Conversation.test.tsx
```diff
@@ -43,19 +43,15 @@ it("collapses read messages, with the exception of the last message in the conve
   )
   await updates(app)
 
-  expect(
-    app
-      .find("Presentable")
-      .filterWhere(n => n.props().presentable.id === "11")
-      .find(Collapse)
-  ).toHaveProp("in", false)
+  expect(app.find({ presentable: { id: "11" } }).find(Collapse)).toHaveProp(
```
hallettj, line 46: I noticed that there were type errors in this test file. I did some research and found this pattern for finding elements by prop values, which is a little more concise.

packages/client/src/DisplayContent.tsx
```diff
@@ -17,71 +19,77 @@ const useStyles = makeStyles(_theme => ({
...
-    )
-  } else {
-    return displayUnknown(
-      { type, subtype, content },
-      clsx(className, classes.body)
-    )
-  }
+  return (
+    <Location>
+      {({ navigate }) => {
```
hallettj, line 34: `Location` is a higher-order component provided by `@reach/router` that provides access to routing objects from context such as the `navigate` function.

packages/client/src/DisplayContent.tsx
```diff
@@ -90,29 +98,43 @@ function displayUnknown(
...
     event.preventDefault()
+
+    const parsed = parseMidUri(target.href)
+    const messageId = parsed && parsed.messageId
+    if (messageId) {
+      navigate(
+        `/accounts/${accountId}/conversations/${encodeURIComponent(messageId)}`
+      )
+      return
+    }
```
hallettj, line 136: If a link can be parsed as a `mid:` URI then we want to navigate to the corresponding conversation in the app. Otherwise we want to open the link in a browser using `openExternal`. For background on `mid:` URIs see https://tools.ietf.org/html/rfc2392

packages/client/src/editor/Editor.tsx
```diff
@@ -1,3 +1,4 @@
+import { idFromHeaderValue, midUri } from "poodle-common/lib/models/uri"
```
hallettj, line 1: Here is an example of importing helper functions from the new common package.

packages/client/src/testing/fixtures.ts
```diff
@@ -57,13 +57,13 @@ export const conversation: graphql.Conversation = {
...
-            contentId: "text"
+              "<CAGM-pNvwffuB_LRE4zP7vaO2noOQ0p0qJ8UmSONP3k8ycyo3HA@mail.gmail.com>",
+            contentId: "replytext"
           },
           resource: {
             messageId:
-              "<CAGM-pNt++x_o=ZHd_apBYpYntkGWOxF2=Q7H-cGEDUoYUzPOfA@mail.gmail.com>",
-            contentId: "text"
+              "<CAGM-pNvwffuB_LRE4zP7vaO2noOQ0p0qJ8UmSONP3k8ycyo3HA@mail.gmail.com>",
+            contentId: "replytext"
```
hallettj, line 66: In principle message IDs and content IDs should be globally unique, so I changed the test fixtures to reflect that.

packages/client/tsconfig.json
```diff
@@ -20,6 +20,9 @@
     "jsx": "preserve",
     "downlevelIteration": true
   },
+  "references": [
+    { "path": "../common" }
+  ],
```
hallettj, line 25: The `references` sections in the client and main package's tsconfigs causes the common package to automatically build when running `tsc --build` or `tsc -b`. For more on that Typescript feature see https://www.typescriptlang.org/docs/handbook/project-references.html

packages/main/package.json
```diff
@@ -7,7 +7,7 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
-    "build": "tsc",
+    "build": "tsc -b",
```
hallettj, line 10: I added a `-b` flag which causes any referenced projects to automatically rebuild. This way the common package will automatically build when we run `yarn start` or `yarn build` in the main package.

packages/main/src/cache/persist.ts
```diff
@@ -1,4 +1,5 @@
 import imap from "imap"
+import { idFromHeaderValue } from "poodle-common/lib/models/uri"
```
hallettj, line 2: Message IDs are strings that come with angle brackets around them. But the brackets are not really part of the ID. `idFromHeaderValue` removes surrounding angle brackets from a string if there are any.

breitman (approved):

PR #28, opened 2019-07-17
star/un-star a conversation

breitman:


hallettj:
I'd like to use a star character that is more colorful, like this one: ⭐

hallettj:
I'm sorry, it's my fault that the Apollo client cache was not updating after mock responses. I explained that when Apollo reads a response from a mutation it can look up a value in the cache with the same type and id and update that value automatically, and I explained that responses must include an `id` property for that to work. What I forgot is that responses also need to have a `__typename` property. You don't need to request `__typename` in operations - it is included automatically in graphql responses. But because the test fixture responses are handwritten we do need to put in `__typename` properties explicitly there.

I pushed a change to add the necessary `__typename` properties, and now you're tests pass. You can remove the second `$.getAccountMock` because the response from `$.flagMock` is sufficient to update client state. I also made a configuration change so that we will see warnings in test output if there are missing `__typename` properties in the future.

packages/client/src/Conversation.tsx
```diff
@@ -125,11 +130,19 @@ export default function Conversation({
     navigate!(`/accounts/${accountId}/dashboard`)
   }
 
+  async function onFlag() {
+    ;(await conversationId) &&
```
hallettj: `conversationId` is not a promise, so this `await` does not add anything. Maybe you meant to `await` the entire expression? As in `await (conversationId && flag({ ... }))`? I'm guessing that you originally wrote `await conversationId && flag({ ... })` but did not realize that `await` has higher precedence than `&&`.

packages/client/src/Conversation.tsx
```diff
@@ -157,12 +170,15 @@ export default function Conversation({
           <IconButton color="inherit" aria-label="archive" onClick={onArchive}>
             <ArchiveIcon />
           </IconButton>
+          <IconButton color="inherit" aria-label="star" onClick={onFlag}>
```
hallettj: Could the `aria-label` switch between `"star"` and `"unstar"`?

packages/client/src/Conversation.tsx
```diff
@@ -157,12 +170,15 @@ export default function Conversation({
...
+            {isStarred ? <StarBorder /> : <StarIcon />}
```
hallettj: I like this icon switch!

packages/client/src/Conversation.tsx
```diff
@@ -157,12 +170,15 @@ export default function Conversation({
...
+          </IconButton>
         </Toolbar>
       </AppBar>
       <CssBaseline />
       <main className={classes.content}>
         <div className={classes.appBarSpacer} />
-        <DisplayErrors results={[archiveResult, setIsReadResult]} />
+        <DisplayErrors results={[archiveResult, setIsReadResult, flagResult]} />
```
hallettj, line 202: Good stuff

packages/client/src/Dashboard.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+import StarIcon from "@material-ui/icons/Star"
+import StarBorder from "@material-ui/icons/StarBorder"
+import * as React from "react"
+import Avatar from "./Avatar"
+import Conversation from "./Conversation"
+import Dashboard from "./Dashboard"
+import { mount, updates } from "./testing"
+import * as $ from "./testing/fixtures"
+
+it("stars a conversation in list view", async () => {
```
hallettj, line 11: fantastic!

packages/main/src/resolvers/conversation.ts
```diff
@@ -50,6 +50,12 @@ export const Conversation: ConversationResolvers = {
     )
   },
 
+  isStarred({ messages }: C.Conversation) {
```
hallettj, line 53: nice

packages/main/src/resolvers/conversation.ts
```diff
@@ -91,6 +97,30 @@ export const ConversationMutations: ConversationMutationsResolvers = {
     return thread
   },
 
+  async flag(_parent, { ids, isFlagged }) {
+    let threads: C.Conversation[] = []
+    ids.forEach(id => {
+      const thread = C.mustGetConversation(id)
+      updateAction(thread.messages, (accountId, box, uids) => {
+        schedule(
+          isFlagged
```
hallettj: I understand the thinking here; but I think that `isFlagged` should reflect the result that you want. So if `isFlagged` is `true` then this resolver should add a star.

packages/main/src/resolvers/conversation.ts
```diff
@@ -91,6 +97,30 @@ export const ConversationMutations: ConversationMutationsResolvers = {
     return thread
   },
 
+  async flag(_parent, { ids, isFlagged }) {
+    let threads: C.Conversation[] = []
+    ids.forEach(id => {
```
hallettj: The more idiomatic way to iterate over an array is `for (const id of ids) { ... }`

packages/client/src/Dashboard.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+      ]
+    }
+  )
+
+  await updates(app)
+  app.find('button[aria-label="star"]').simulate("click")
+  await updates(app, 10)
+
+  expect(app.find(StarIcon)).toExist()
+})
```
hallettj, line 178: These are good tests! I would like to add some tests for starring and unstarring multiple conversations at once. I noticed that when multiple conversations are selected if only some are starred then clicking the star action removes the stars. The behavior is nicely consistent - but I would like it to be flipped so that clicking the star action adds stars unless all of the selected conversations are starred.

packages/client/src/Dashboard.tsx
```diff
@@ -109,9 +111,23 @@ export default function Dashboard({ accountId, navigate }: Props) {
...
+  selected.forEach(select => {
+    conversations &&
+      conversations.forEach(conversation => {
+        if (conversation.id === select) {
+          if (conversation.isStarred) {
+            isStarred = true
+          }
+        }
+      })
+  })
```
hallettj: Instead of `forEach` would you use `reduce`? We can work on that change together if you want.

packages/client/src/Dashboard.tsx
```diff
@@ -243,9 +274,12 @@ function SelectedActionsBar({
           <IconButton aria-label="archive" onClick={onArchive}>
             <ArchiveIcon />
           </IconButton>
+          <IconButton aria-label="star" onClick={onFlag}>
```
hallettj: Would you also make this aria-label switch between "star" and "unstar"?

packages/client/src/Dashboard.tsx
```diff
@@ -362,9 +398,10 @@ function ConversationRow({
           />
         )}
       </ListItemAvatar>
+
       <ListItemText
         id={rowId}
-        primary={subject || "[no subject]"}
+        primary={(isStarred ? "⭐ " : "") + (subject || "[no subject]")}
```
hallettj, line 416: Hey this is good

packages/main/jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   preset: "ts-jest",
   runner: "@jest-runner/electron/main",
-  setupFiles: ["./jest.setup.js"],
+  setupFilesAfterEnv: ["./jest.setup.js"],
```
hallettj, line 4: I made a change a while ago that I thought would increase test timeouts so that we don't get these intermittent failures. But apparently I did it wrong. I pushed this change to your branch to fix that.

packages/main/schema/schema.graphql
```diff
@@ -38,6 +39,7 @@ type ConversationSearchResult {
 
 type ConversationMutations {
   archive(id: ID!): Conversation!
+  flag(ids: [ID!]!, isFlagged: Boolean!): [Conversation!]!
```
hallettj, line 42: nifty

packages/main/src/models/conversation.ts
```diff
@@ -109,6 +109,9 @@ export function getPresentableElements({
         isRead: cache
           .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
           .includes("\\Seen"),
+        isStarred: cache
+          .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
+          .includes("\\Flagged"),
```
hallettj: This will be handy when we display stars per presentable. But I would like this to be computed differently: if you star a message the star should remain when the message is edited. And it it should be possible to a remove a star at the presentable level even if the presentable has been edited, which is only possible if the star is displayed. In other words a presentable should be considered starred if any of its revisions is starred.

This is the sort of behavior that is especially useful to write tests for - what happens to state when these things happen? It's kinda like board game rules: you don't want situations to come up where the game rules are genuinely ambiguous on what should happen. A game designer should try to anticipate that kind of thing, and make sure that the game rules are comprehensive enough to cover all situations. In an application tests should be comprehensive enough to specify what should happen in situations where a feature might reasonably be implemented in different ways.

But like I said, this computation is not currently used because there is no per-presentable star control yet. I added a note in "This Week" to add that feature.

packages/main/src/queue/index.ts
```diff
@@ -246,6 +246,72 @@ const handlers = {
         sync(accountId, connectionManager)
       )
     }
+  }),
+
+  flag: handler({
```
hallettj: The `flag` and `unflag` handlers are nearly identical. They could be combined into one `setFlagged` handler with a boolean parameter.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -411,6 +411,62 @@ describe("when querying conversations", () => {
     })
   })
 
+  it("stars a conversation", async () => {
```
hallettj, line 414: More good tests!

hallettj (changes requested):
This is great work! You have implemented a feature that touches nearly every part of the tech stack for this app, and you have shown good initiative.

packages/client/src/Conversation.tsx
```diff
@@ -167,12 +168,32 @@ export default function Conversation({
           >
             {subject}
           </Typography>
-          <IconButton color="inherit" aria-label="archive" onClick={onArchive}>
-            <ArchiveIcon />
-          </IconButton>
-          <IconButton color="inherit" aria-label="star" onClick={onFlag}>
-            {isStarred ? <StarBorder /> : <StarIcon />}
-          </IconButton>
+          <Tooltip
```
hallettj, line 171: Good stuff

packages/client/src/Dashboard.test.tsx
```diff
@@ -55,12 +55,132 @@ it("un-stars a conversation in list view", async () => {
...
+      .find(ListItemText)
+      .filterWhere(node => node.prop("id") === "conversation-row-1")
+  ).not.toIncludeText("⭐ ")
+  expect(
+    app
+      .find(ListItemText)
+      .filterWhere(node => node.prop("id") === "conversation-row-2")
+  ).not.toIncludeText("⭐ ")
+})
+
```
hallettj, line 135: Good tests!

packages/client/src/Dashboard.tsx
```diff
@@ -115,18 +116,27 @@ export default function Dashboard({ accountId, navigate }: Props) {
...
-        if (conversation.id === select) {
-          if (conversation.isStarred) {
-            isStarred = true
-          }
-        }
-      })
-  })
+  //if not all selected star, we want to star instead of unstar
+  if (isStarred && selected.length > 1) {
+    isStarred =
```
hallettj: Assigning `isStarred` twice is a [code smell](https://martinfowler.com/bliki/CodeSmell.html). The logic for this second assignment is correct - using `every` instead of `reduce` covers the cases that you need. The first assignment is redundant. You can delete the first assignment, and delare `isStarred` with `const`.

packages/client/src/Dashboard.tsx
```diff
@@ -271,12 +281,27 @@ function SelectedActionsBar({
...
         <Toolbar className={classes.toolbar}>
           <span className={classes.title} />
-          <IconButton aria-label="archive" onClick={onArchive}>
-            <ArchiveIcon />
-          </IconButton>
-          <IconButton aria-label="star" onClick={onFlag}>
-            {isStarred ? <StarBorder /> : <StarIcon />}
-          </IconButton>
+          <Tooltip
+            title="Archive Selected Conversation"
```
hallettj, line 272: These tooltips are great! But please pluralize "Conversations" for the list view actions.

packages/client/src/testing/fixtures.ts
```diff
@@ -153,7 +153,10 @@ export const archiveMock = {
 export const flagMock = {
   request: {
     query: graphql.FlagDocument,
-    variables: { conversationIDs: [conversation.id], isFlagged: false }
+    variables: {
+      conversationIDs: [conversation.id],
+      isFlagged: conversation.isStarred
```
hallettj: It's a little weird to reference `conversation.isStarred` for the expected variable value, and to use a literal value `true` for the `isStarred` value in the mock response. Either both values should be `true`, or both should be `false`.

Something you can do that might make testing smoother is to make `flagMock` a function instead of a plain value. It could accept an argument for `isStarred` to make it easier to use the mock for both the star and unstar cases.

packages/main/src/queue/index.ts
```diff
@@ -248,36 +248,54 @@ const handlers = {
     }
   }),
 
-  flag: handler({
-    enqueue(params: { accountId: ID; box: { name: string }; uids: number[] }) {
-      cache.addFlag({ ...params, flag: "\\Flagged" })
+  setFlagged: handler({
```
hallettj, line 251: Good stuff. Would you remove the `unFlag` handler now that it is not used anymore?

packages/main/src/queue/index.ts
```diff
@@ -248,36 +248,54 @@ const handlers = {
...
+    enqueue(params: {
+      accountId: ID
+      box: { name: string }
+      uids: number[]
+      isFlagged: boolean
+    }) {
+      params.isFlagged
+        ? cache.delFlags({ ...params, flags: ["\\Flagged"] })
+        : cache.addFlag({ ...params, flag: "\\Flagged" })
```
hallettj: This is still backward. If `isFlagged` is `true` I think that means that we are instructing the service to make the conversation flagged. Don't think of `isFlagged` as the prior state of the conversation, think of it as the end result.

packages/main/src/resolvers/conversation.ts
```diff
@@ -99,25 +99,20 @@ export const ConversationMutations: ConversationMutationsResolvers = {
...
-              })
-            : actions.flag({
-                accountId: String(accountId),
-                box,
-                uids
-              })
+          actions.setFlagged({
+            accountId: String(accountId),
+            box,
+            uids,
```
hallettj: I should have mentioned earlier - this will flag every message in the conversation. I want to only flag the last message. It is sufficient to take the last value in the `uids` list. A convenient take a list and produce a new list that only keeps the last value is `uids.slice(-1)`.

hallettj (changes requested):
There is lots of good progress here. Please remember to either remove or update the `isStarred` calculation in `models/conversation.ts`.

packages/main/src/models/conversation.ts
```diff
@@ -109,9 +109,7 @@ export function getPresentableElements({
         isRead: cache
           .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
           .includes("\\Seen"),
-        isStarred: cache
-          .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
-          .includes("\\Flagged"),
+        isStarred: cache.getFlags(message.id).includes("\\Flagged"),
```
hallettj: This still does not work. In this context `message` is the first revision of a presentable. If the presentable has been edited it is possible that one of the edit messages has been flagged, and in that case `isStarred` should be `true`. But this line does not check flags on edit messages. I recommend removing this property from this pull request, and getting back to it in the per-presentable starring feature.

packages/main/src/resolvers/conversation.ts
```diff
@@ -102,11 +102,12 @@ export const ConversationMutations: ConversationMutationsResolvers = {
     for (const id of ids) {
       const thread = C.mustGetConversation(id)
       updateAction(thread.messages, (accountId, box, uids) => {
+        console.log(uids)
```
hallettj: Please remove this `console.log` line.

hallettj (changes requested):
More good progress! But there are some outstanding comments from my previous review.

PR #30, opened 2019-07-23
star individual presentables in conversation view

breitman:


packages/client/src/Conversation.tsx
```diff
@@ -256,6 +235,17 @@ function Presentable({
     setAnchorEl(null)
   }
 
+  async function onFlag() {
+    await (conversationId &&
+      flag({
+        variables: {
+          conversationId: conversationId!,
```
hallettj: Minor detail, but I think that the non-null assertion is not necessary here because you checked that `conversationId` is truthy earlier in this expression.

packages/client/src/Conversation.tsx
```diff
@@ -276,6 +266,22 @@ function Presentable({
             >
               <MoreVertIcon />
             </IconButton>
+            <Tooltip
+              title={
+                (presentable.isStarred ? "Unstar " : "Star ") + "Conversation"
+              }
+              enterDelay={500}
```
hallettj: Since we are going to want to use the same `enterDelay` and `leaveDelay` everywhere it would be useful at some point to define a custom Tooltip component that sets these values by default. That way if we want to adjust those parameters we can do so in one place.

packages/client/src/Conversation.tsx
```diff
@@ -276,6 +266,22 @@ function Presentable({
...
+              leaveDelay={200}
+            >
+              <IconButton
+                aria-controls={`star-${presentable.id}`}
```
hallettj: Where is the element with this id? `aria-controls` is used when clicking one element causes a change to a different element (such as causing a menu to appear) and we use the `id` of that other element as the value of `aria-controls`.

packages/client/src/Dashboard.test.tsx
```diff
@@ -133,25 +133,28 @@ it("unstars selected conversations when all are starred", async () => {
   ).not.toIncludeText("⭐ ")
 })
 
-it("stars a conversation in conversation view", async () => {
+it("stars a presentable in conversation view", async () => {
```
hallettj: These tests are great, but they should be in `Conversation.test.tsx`.

packages/client/src/Dashboard.test.tsx
```diff
@@ -133,25 +133,28 @@ it("unstars selected conversations when all are starred", async () => {
...
+        $.getConversationMock,
+        $.flagPresentableMock({ isFlagged: true, presentableId: "11" })
+      ]
     }
   )
 
   await updates(app)
-  app.find('button[aria-label="star"]').simulate("click")
-  await updates(app, 10)
+  app.find('button[aria-controls="star-11"]').simulate("click")
```
hallettj: An alternative way to find this element if you don't have `aria-controls` available is,

```ts
app.find({ presentable: { id: "11" }}).find(StarIcon).simulate("click")
```

packages/client/src/Dashboard.test.tsx
```diff
@@ -133,25 +133,28 @@ it("unstars selected conversations when all are starred", async () => {
...
+  await updates(app, 100)
 
   expect(app.find(StarBorder)).toExist()
```
hallettj: I think you want to check for a `StarIcon`?

packages/client/src/Dashboard.test.tsx
```diff
@@ -133,25 +133,28 @@ it("unstars selected conversations when all are starred", async () => {
...
+        $.flagPresentableMock({ isFlagged: true, presentableId: "11" })
+      ]
     }
   )
 
   await updates(app)
-  app.find('button[aria-label="star"]').simulate("click")
-  await updates(app, 10)
+  app.find('button[aria-controls="star-11"]').simulate("click")
+  await updates(app, 100)
```
hallettj: You don't need such a long delay here. I think this will work if you don't specify any delay.

packages/client/src/Dashboard.test.tsx
```diff
@@ -162,17 +165,29 @@ it("un-stars a conversation in conversation view", async () => {
         {
           ...$.getConversationMock,
           result: {
-            data: { conversation: { ...$.conversation, isStarred: true } }
+            data: {
```
hallettj: good stuff

packages/client/src/documents/operations.graphql
```diff
@@ -210,6 +211,30 @@ mutation flag($conversationIDs: [ID!]!, $isFlagged: Boolean!) {
...
+    ) {
+      id
+      date
+      from {
+        host
+        mailbox
+        name
+      }
+      isRead
+      isStarred
```
hallettj, line 133: When getting data back from a mutation we only need to request properties that might have changed, and the ids to help Apollo to match changes with cached records. In this case that means the presentable `isStarred` and `id`, and the conversation `isStarred` and `id`, because changing the `isStarred` status of a presentable can also change the status of the parent conversation. In the future we might want to indicate the starred state of presentables and the conversation in one view, and it will be important to have the Apollo cache updated for both.

In this case you fetch the updated `isStarred` property at the conversation level, but not at the presentable level.

We also want to make sure that if you star or unstar a presentable in conversation view and then switch to dashboard view that the conversation shows the correct starred status there.

packages/client/src/testing/fixtures.ts
```diff
@@ -178,6 +180,33 @@ export function flagMock({
   }
 }
 
+export function flagPresentableMock({
```
hallettj, line 283: Nice! But it looks like need to set the `isStarred` properties in the result data.

packages/main/schema/schema.graphql
```diff
@@ -40,6 +40,11 @@ type ConversationSearchResult {
 type ConversationMutations {
   archive(id: ID!): Conversation!
   flag(ids: [ID!]!, isFlagged: Boolean!): [Conversation!]!
+  flagPresentable(
+    id: ID!
+    conversationId: ID!
+    isFlagged: Boolean!
+  ): Conversation!
```
hallettj, line 48: good stuff

packages/main/src/models/conversation.ts
```diff
@@ -109,6 +109,7 @@ export function getPresentableElements({
         isRead: cache
           .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
           .includes("\\Seen"),
+        isStarred: cache.getFlags(message.id).includes("\\Flagged"),
```
hallettj: You need to check whether any revision includes a `"\\Flagged"` flag, not just the first message. At this point in the code every element in the `resources` collection has a `revision` property, which has a reference to the message that provided that revision.

I'm sorry about the weird overlap of the words "revision" and "resource"! For elements in the `resources` collection the `resource` property is a reference to the first revision which gives subsequent revisions a consistent way to identify what they have edited, and the `revision` property is a reference to the edit message that produced the new revision.

packages/main/src/resolvers/conversation.ts
```diff
@@ -116,6 +116,26 @@ export const ConversationMutations: ConversationMutationsResolvers = {
     return threads
   },
 
+  async flagPresentable(_parent, { conversationId, id, isFlagged }) {
+    const thread = C.mustGetConversation(conversationId)
+    for (const message of thread.messages) {
+      if (String(message.id) === id) {
```
hallettj: This will probably work when using Poodle. But this will not properly unstar edit messages that were starred from Gmail. The id of the presentable happens to be the cache id of the message that provides the first revision of the presentable. But there is a cleaner way to do this: call `getPresentableElements(thread)`, find the presentable with the matching id, and get the list of messages making up that presentable from the presentable object itself. As I said before that involves adding a list of `revisions` to the `Presentable` type.

For purposes of interoperation with other email clients I think it would make most sense to add a flag the latest edit message instead of the first revision. When removing flags it is important to remove from all messages in the presentable revision history that have flags set.

I'd like to see tests to make sure that starring and unstarring works with presentables that have been edited.

packages/main/src/resolvers/conversation.ts
```diff
@@ -205,7 +225,7 @@ function setIsRead(messages: cache.Message[], isRead: boolean) {
  * Groups messages by account and box as a convenience for dispatching IMAP
  * requests.
  */
-function updateAction(
+export function updateAction(
```
hallettj: Does this need to be exported?

hallettj (changes requested):

packages/client/src/Conversation.tsx
```diff
@@ -274,11 +273,14 @@ function Presentable({
               leaveDelay={200}
             >
               <IconButton
-                aria-controls={`star-${presentable.id}`}
                 aria-label={presentable.isStarred ? "unstar" : "star"}
                 onClick={onFlag}
               >
-                {presentable.isStarred ? <StarBorder /> : <StarIcon />}
+                {presentable.isStarred ? (
+                  <StarIcon style={{ fill: "gold" }} />
```
hallettj, line 274: oh, shiny

packages/client/src/documents/operations.graphql
```diff
@@ -196,17 +196,7 @@ mutation flag($conversationIDs: [ID!]!, $isFlagged: Boolean!) {
   conversations {
     flag(ids: $conversationIDs, isFlagged: $isFlagged) {
```
hallettj, line 106: Just like changing the flag state of a presentable can change the flag state of the conversation, changing the flag state of a conversation can change the flag state of its presentables. The fields selected in the `flag` operation should be the same as those selected in the `flagPresentable` operation.

packages/client/src/testing/fixtures.ts
```diff
@@ -150,6 +150,41 @@ export const archiveMock = {
   }
 }
 
+// export function editMock({
```
hallettj: Looks like you decided not to use this mock? Would you remove the commented code?

packages/client/src/testing/fixtures.ts
```diff
@@ -200,7 +235,19 @@ export function flagPresentableMock({
...
         conversations: {
           __typename: "ConversationMutations",
-          flagPresentable: conversation
+          flagPresentable: {
+            ...conversation,
+            isStarred: isFlagged,
+            presentableElements: [
+              {
+                ...conversation.presentableElements[0],
+                isStarred: isFlagged
```
hallettj: This is good; but since you have a value for `presentableId` I think it would be useful to change the `isStarred` state on the presentable with a matching id.

packages/main/src/models/conversation.ts
```diff
@@ -12,6 +12,18 @@ export interface Conversation {
...
+  id: string
+  isRead: boolean
+  isStarred: boolean
+  contents: Content[]
+  date: string
+  from: Address
+  editedAt?: string
+  editedBy?: Address
+  revisions: cache.Message[]
+}
```
hallettj, line 21: It was good thinking to create this type here. Another way to do this which would keep this `Presentable` type in sync with the one from GraphQL would be:

```ts
import { Presentable as GeneratedPresentable } from "../generated/graphql"

export interface Presentable extends GeneratedPresentable {
  revisions: cache.Message[]
}
```

packages/main/src/models/conversation.ts
```diff
@@ -109,14 +121,20 @@ export function getPresentableElements({
         isRead: cache
           .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
           .includes("\\Seen"),
-        isStarred: cache.getFlags(message.id).includes("\\Flagged"),
+        isStarred: resources.some(r =>
+          cache.getFlags(r.revision.message.id).includes("\\Flagged")
+        ),
```
hallettj, line 122: Great!

packages/main/src/models/conversation.ts
```diff
@@ -109,14 +121,20 @@ export function getPresentableElements({
...
           .valueSeq()
           .map(getPresentableContent)
           .toArray(),
         date: message.date,
         from: cache.getParticipants(message.id, "from")[0],
         editedAt: latestEdit && latestEdit.revision.message.date,
+        revisions: resources
+          .valueSeq()
+          .flatMap(r => [r.revision.message])
+          .toArray(),
```
hallettj, line 133: Good stuff. But I realize that although we only need messages at the moment, it would be highly useful for the future if the `revisions` property includes the entire revision object - both message and part.

Also this use of `flatMap` could be simplified to `.map(r => r.revision.message)`. Calling `xs.flatMap(fn)` is the same as calling `xs.map(fn).flatten()`. In this case after `valueSeq` you get a sequence that looks roughly like this:

```ts
[{ message: m1, part: p1 }, { message: m2, part: p2 }]
```

If we think if the `flatMap` call as two steps (`map` and `flatten`) then after the `map` your sequence looks like this:

```ts
[[m1], [m2]]
```

`flatten` turns a list of lists into one flat list; so after `flatten` the result looks like this:

```ts
[m1, m2]
```

The only reason that flattening is necessary is because you wrap each message in a new array in the `flatMap` callback. If you skip the array wrapping you can use `map` directly and skip a step.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -616,6 +618,92 @@ describe("when querying conversations", () => {
     })
   })
 
+  it("stars an edited message", async () => {
```
hallettj, line 621: Perfect!

packages/main/src/resolvers/conversation.ts
```diff
@@ -133,7 +134,7 @@ export const ConversationMutations: ConversationMutationsResolvers = {
       }
     }
 
-    return C.mustGetConversation(conversationId)
+    return thread
```
hallettj: Why did you change this? If you don't re-fetch the conversation from the cache then the client will not get the updated flag state.

packages/main/src/resolvers/conversation.ts
```diff
@@ -118,9 +118,10 @@ export const ConversationMutations: ConversationMutationsResolvers = {
 
   async flagPresentable(_parent, { conversationId, id, isFlagged }) {
     const thread = C.mustGetConversation(conversationId)
-    for (const message of thread.messages) {
-      if (String(message.id) === id) {
-        updateAction([message], (accountId, box, uids) => {
+    const presentables = C.getPresentableElements(thread)
+    for (const presentable of presentables) {
+      if (presentable.id === id) {
+        updateAction(presentable.revisions, (accountId, box, uids) => {
```
hallettj: This is better! But when adding a flag I would like to flag the last revision instead of all revisions. When removing flags we should filter to revisions that are flagged, and just remove flags from those.

hallettj (changes requested):

packages/client/src/Tooltip.tsx
```diff
@@ -0,0 +1,16 @@
+import { Tooltip as TTip } from "@material-ui/core"
+import React, { ReactElement } from "react"
+
+export default function Tooltip({
```
hallettj, line 4: Nifty!

There is a more generalized way to wrap a component and pass through all of the props that it accepts. I looked at the type definition for `Tooltip` and saw that its props type is exported as `TooltipProps`. You want to accept any props that `Tooltip` can accept, but you want to pull out `enterDelay` and `leaveDelay` to set your own default values if the caller does not provide override values. The remaining props should be passed through. You can do that like this:

```ts
import { Tooltip as TTip } from "@material-ui/core"
import { TooltipProps } from "@material-ui/core/Tooltip"
import React from "react"

export default function Tooltip({
  enterDelay = 500,
  leaveDelay = 200,
  ...rest
}: TooltipProps) {
  return <TTip enterDelay={enterDelay} leaveDelay={leaveDelay} {...rest} />
}
```

Remember that `children` is just another prop, and it is passed through with everything else via `rest`.

Assigning default values when destructuring allows the caller to override `enterDelay` or `leaveDelay` if they want to.

packages/client/src/testing/fixtures.ts
```diff
@@ -238,15 +203,14 @@ export function flagPresentableMock({
...
-                ...conversation.presentableElements[0],
-                isStarred: isFlagged
-              },
-              {
-                ...conversation.presentableElements[1]
+            presentableElements: conversation.presentableElements.map(
+              presentable => {
+                if (presentable.id === presentableId) {
+                  return { ...presentable, isStarred: isFlagged }
+                }
```
hallettj, line 310: good stuff!

packages/main/src/models/conversation.ts
```diff
@@ -12,16 +16,8 @@ export interface Conversation {
...
-  id: string
-  isRead: boolean
-  isStarred: boolean
-  contents: Content[]
-  date: string
-  from: Address
-  editedAt?: string
-  editedBy?: Address
-  revisions: cache.Message[]
+export interface Presentable extends GeneratedPresentable {
```
hallettj, line 19: nice

hallettj (approved):
Looks great! I left some feedback on the implementation of the new `Tooltip`, but that is a future-reference thing, and is not important to change for this feature.

You'll need to merge changes from master, fix some conflicts, and make sure that the tests pass.

PR #31, opened 2019-07-24
close add account modal when clicking "add account" button in account view

breitman:


packages/client/src/Accounts.tsx
```diff
@@ -78,7 +78,7 @@ export default function Accounts(_props: RouteComponentProps) {
   const [open, setOpen] = React.useState(false)
 
   const handleClick = () => {
-    open ? setOpen(false) : setOpen(true)
+    setOpen(!open)
```
hallettj, line 81: Yep, this is better.

packages/client/src/Accounts.tsx
```diff
@@ -90,6 +90,7 @@ export default function Accounts(_props: RouteComponentProps) {
     try {
       await addAccount({ variables: { email: emailValue } })
       setEmailValue("")
+      setOpen(false)
```
hallettj, line 93: Great, thanks!

hallettj (approved):

PR #32, opened 2019-07-24
limit rate of UI updates while syncing

breitman:


packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,23 @@
+import debounce from "./debounce"
+
+jest.useFakeTimers()
+
+describe("debounce", () => {
+  let func: jest.Mock
+  let debouncedFunc: Function
+
+  beforeEach(() => {
+    func = jest.fn()
```
hallettj, line 10: good thinking

packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  })
+
+  it("executes after 3 sec delay when called multiple times", () => {
+    for (let i = 0; i < 300; i++) {
+      debouncedFunc()
+    }
+
+    jest.runAllTimers()
+
+    expect(func).toBeCalledTimes(1)
```
hallettj, line 24: nice

packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+  })
+})
```
hallettj, line 56: I would like a test for this scenario: every time `debouncedFunc` is called it is called with a different argument. Verify that in any conditions `func` is eventually called with the same argument as the _last_ call to `debouncedFunc`.

I would also like to test that if `debouncedFunc` is called many times over a period longer than 3 seconds that `func` is called every 3 seconds.

packages/client/src/debounce.tsx
```diff
@@ -0,0 +1,18 @@
+export default function debounce(
+  func: Function,
```
hallettj, line 2: The type for `func` should be more precise. In particular `debounce` should return a function with the same signature as `func`. That is something we can work on together.

packages/client/src/debounce.tsx
```diff
@@ -0,0 +1,18 @@
...
+  wait: number,
+  immediate: boolean
+) {
+  let timeout: any
+  return function(this: any, ...args: any[]) {
+    const context = this
```
hallettj: If you use an arrow function for `later` then you wouldn't need the intermediate `context` variable.

packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,23 @@
+import debounce from "./debounce"
+
+jest.useFakeTimers()
```
hallettj, line 3: nice

packages/client/src/index.tsx
```diff
@@ -31,11 +32,20 @@ const client = new ApolloClient({
...
-ipcRenderer.on("message_updates", () => {
-  if (client.queryManager) {
-    client.queryManager.reFetchObservableQueries()
-  }
-})
+
+ipcRenderer.on(
+  "message_updates",
+  debounce(
+    function() {
```
hallettj: For inline callbacks I usually prefer arrow functions.

packages/client/src/index.tsx
```diff
@@ -31,11 +32,20 @@ const client = new ApolloClient({
...
+      if (client.queryManager) {
+        client.queryManager.reFetchObservableQueries()
+      }
+    },
+    3000,
+    false
```
hallettj: I think this use should be immediate.

hallettj (changes requested):

packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  it("calls func every 3 seconds over a period longer than 3 seconds", () => {
+    debouncedFunc()
+    expect(func).toBeCalledTimes(1)
+
+    for (let i = 0; i < 10; i++) {
+      debouncedFunc()
+      jest.advanceTimersByTime(3000)
+    }
+
+    expect(func).toBeCalledTimes(10)
```
hallettj, line 36: This does not quite test what we want because advancing the timers after every call gives time for `debounce`'s timeout to clear. We want to test that `func` is called every three seconds while `debouncedFunc` is called multiple times during any given three-second window.

I'm calling this out because there is a problem with the implementation of `debounce`. This test and the next test should be adjusted until they both fail.

packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  it("eventually calls func with the same argument as the last call to debouncedFunc", () => {
+    let i
+    let args: number[] = []
+
+    for (i = 0; i < 5; i++) {
+      const rando = Math.floor(Math.random() * 6 + 1)
+      debouncedFunc(rando)
+      args.push(rando)
+      jest.advanceTimersByTime(3000)
```
hallettj, line 47: This is close to what I want. But because this test advances timers after every call we don't test cases where a call is delayed to the next three-second window.

This test should call `debouncedFunc` multiple times before advancing timers, then verify that `func` was called with the last argument after advancing timers.

packages/client/src/debounce.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+      debouncedFunc()
+      jest.advanceTimersByTime(3000)
+    }
+
+    expect(func).toBeCalledTimes(10)
+  })
+
+  it("eventually calls func with the same argument as the last call to debouncedFunc", () => {
+    let i
+    let args: number[] = []
```
hallettj, line 41: You only really care about the last argument, so you don't need an array here. If you just track the latest argument then `i` can be scoped to only the `for` block.

hallettj (changes requested):

hallettj:
I'm sorry, I think I asked for the wrong thing. I asked for a "debounce" when I meant to ask for a "throttle". The function that you wrote behaves correctly for a debounce where receiving many repeated calls schedules the underlying function, and keeps pushing off that call until there is a pause in incoming calls.

PR #33, opened 2019-07-24
show labels in list view

breitman:


packages/client/src/Dashboard.tsx
```diff
@@ -371,11 +373,25 @@ function ConversationRow({
...
+    id,
+    isRead,
+    snippet,
+    subject,
+    labels,
+    isStarred
+  } = conversation
+
+  let newLabels = labels
+  newLabels = newLabels && newLabels.map(label => label.slice(1))
```
hallettj: Declarations with `let` are often a code smell. In this case you can combine this into one line and declare with `const`.

Another issue is that `newLabels` is not a descriptive name. Maybe the name could be `labelsForDisplay`?

The `slice` call removes the leading backslash, right? The problem is that custom labels do not begin with a backslash. (See https://developers.google.com/gmail/imap/imap-extensions#access_to_gmail_labels_x-gm-labels) A more robust solution would be to use the `replace` method with a regular expression argument. But this is display logic, and I would rather put display logic as late as possible at the point where a value is actually displayed, which would be in `DisplayLabel`.

packages/client/src/Dashboard.tsx
```diff
@@ -436,8 +452,63 @@ function ConversationRow({
           secondary: classes.snippetText
         }}
       />
+      {newLabels &&
+        newLabels.map(label => (
+          <DisplayLabel key={labelId + count++} label={label} />
```
hallettj: I don't like the use of a `count` variable that is scoped over the entire component. If you want an index for each element when iterating over an array then use the second parameter of the `map` callback:

```ts
newLabels.map((label, index) => (
  <DisplayLabel key={labelId + index} label={label} />
))
```

But using the array index as the `key` prop should be a last resort - it's better to use a key that is intrinsically linked to the thing that is being displayed. In this case the key could be the value of the `label` variable itself.

It looks like you incorporated the conversation id into the key to avoid collisions with keys in other conversation rows. That isn't necessary: a `key` prop only needs to be unique within the same list. So as long as the conversation does not have two labels with the same text then you can use the label text by itself as the key.

packages/client/src/Dashboard.tsx
```diff
@@ -436,8 +452,63 @@ function ConversationRow({
...
+        ))}
     </ListItem>
   )
 }
 
+function DisplayLabel({ label }: { label: string }) {
```
hallettj, line 456: This is good use of a helper component!

hallettj (changes requested):
This is looking good! I think there is an issue preventing all labels from syncing at the moment which makes it difficult to test handling of leading backslashes fully. But that should not be a blocker for this feature.

packages/client/src/Dashboard.tsx
```diff
@@ -452,15 +447,15 @@ function ConversationRow({
...
-          <DisplayLabel key={labelId + count++} label={label} />
-        ))}
+      {labels &&
+        labels.map(label => <DisplayLabel key={label} label={label} />)}
     </ListItem>
   )
 }
 
 function DisplayLabel({ label }: { label: string }) {
+  const labelForDisplay = label.replace(/\\/, "")
```
hallettj: Yes, much better! There is one tweak to be made: this replacement should only be made if the backslash is the first character in the label string. You can restrict a regular expression pattern to the beginning of a string using the `^` anchor:

```ts
label.replace(/^\\/, "")
```

hallettj (changes requested):
Good stuff! I asked for one small change. After that please go ahead and merge this one.

PR #35, opened 2019-07-26
remember search state via URL

breitman:


packages/client/src/App.tsx
```diff
@@ -1,21 +1,24 @@
...
 export default function App() {
   return (
-    <Router>
-      <Init path="/" />
-      <Accounts path="accounts" />
-      <Compose path="accounts/:accountId/compose" />
-      <Conversation path="accounts/:accountId/conversations/:conversationId" />
-      <Dashboard path="accounts/:accountId/dashboard/" />
-    </Router>
+    <QueryParamProvider reachHistory={globalHistory}>
```
hallettj, line 13: good stuff

packages/client/src/Dashboard.tsx
```diff
@@ -120,14 +121,23 @@ const useStyles = makeStyles(theme => ({
 export default function Dashboard({ accountId, navigate }: Props) {
   const classes = useStyles()
   const [open, setOpen] = React.useState(false)
-  const [isSearching, setIsSearching] = React.useState(false)
-  const [searchQuery, setSearchQuery] = React.useState("")
+  // const [isSearching, setIsSearching] = React.useState(false)
```
hallettj: Please remove commented-out code. We can always look back at revision history if we need to see what was there.

packages/client/src/Conversation.tsx
```diff
@@ -155,7 +155,7 @@ export default function Conversation({
             edge="start"
             color="inherit"
             aria-label="close view"
-            onClick={() => navigate!(`/accounts/${accountId}/dashboard`)}
+            onClick={() => window.history.back()} //navigate!(`/accounts/${accountId}/dashboard`)}
```
hallettj: Please remove commented-out code.

packages/client/src/Dashboard.tsx
```diff
@@ -120,14 +121,23 @@ const useStyles = makeStyles(theme => ({
...
+
+  const [query, setQuery] = useQueryParams({
+    searchQuery: StringParam,
+    isSearching: BooleanParam
+  })
+  const { searchQuery, isSearching } = query
+
+  // const [searchQuery, setSearchQuery] = React.useState("")
```
hallettj: more commented-out code

packages/client/src/Dashboard.tsx
```diff
@@ -120,14 +121,23 @@ const useStyles = makeStyles(theme => ({
...
   const [open, setOpen] = React.useState(false)
-  const [isSearching, setIsSearching] = React.useState(false)
-  const [searchQuery, setSearchQuery] = React.useState("")
+  // const [isSearching, setIsSearching] = React.useState(false)
+
+  const [query, setQuery] = useQueryParams({
+    searchQuery: StringParam,
+    isSearching: BooleanParam
+  })
+  const { searchQuery, isSearching } = query
```
hallettj: Instead of checking for `undefined` later, you can assign default values in a destructuring assignment, like this:

```ts
const { searchQuery = "", isSearching = false } = query
```

An even more terse option which would avoid the intermediate `query` variable is nested destructuring:

```ts
const [{ searchQuery = "", isSearching = false }, setQuery] = useQueryParams({
  searchQuery: StringParam,
  isSearching: BooleanParam
})
```

I think the names `query` and `setQuery` are not descriptive. I would call them `searchParams` and `setSearchParams`.

But that still leaves an issue: `searchQuery` and `isSearching` are independent. Using a single hook for both pushes you into wanting to update both together. I think it would be better to use two hooks so that these values can be set separately as they were when they were backed by `useState`:

```ts
  const [searchQuery = "", setSearchQuery] = useQueryParam(
    "searchQuery",
    StringParam
  )
  const [isSearching = false, setIsSearching] = useQueryParam(
    "isSearching",
    BooleanParam
  )
```

packages/client/src/Dashboard.tsx
```diff
@@ -120,14 +121,23 @@ const useStyles = makeStyles(theme => ({
...
+
+  // const [searchQuery, setSearchQuery] = React.useState("")
   const getAccountResult = graphql.useGetAccountQuery({
     variables: { accountId: accountId! }
   })
-  const skipSearch = !isSearching || searchQuery.length < 3
+  const skipSearch =
+    !isSearching || (searchQuery ? searchQuery.length < 3 : false)
```
hallettj: The ternary could be shortened to `!!searchQuery && searchQuery.length < 3`. But that is a matter of taste - it is not necessarily better.

But as I said above it would be cleaner to assign `searchQuery` so that it is always a `string`.

packages/client/src/Dashboard.tsx
```diff
@@ -120,14 +121,23 @@ const useStyles = makeStyles(theme => ({
...
+
   const searchResult = graphql.useSearchConversationsQuery({
-    variables: { accountId: accountId!, query: searchQuery },
+    variables: { accountId: accountId!, query: searchQuery || "" },
```
hallettj: The `|| ""` would be redundant if `searchQuery` is always a `string`.

packages/client/src/Dashboard.tsx
```diff
@@ -165,17 +175,19 @@ export default function Dashboard({ accountId, navigate }: Props) {
         />
       ) : isSearching ? (
         <SearchBar
-          onChange={setSearchQuery}
+          onChange={(q: string) =>
+            setQuery({ searchQuery: q, isSearching: true }, "replace")
+          }
           onClose={() => {
-            setIsSearching(false)
+            setQuery({ searchQuery: "", isSearching: false }, "replace")
```
hallettj: I would rather not unset the search query when the search input is closed. If the user closes search, and opens it again I want the query and search results that they previously entered to still be there.

hallettj (changes requested):

hallettj:
I'm sorry about the test failures - those are my fault.

I changed the "base" of this pull request to be `jh/search` so that the listed changes are only those from your changes. We'll change the base back to `master` after `jh/search` is merged.

hallettj:
@breitman I updated and merged the search branch, which means that there are some merge conflicts that need to be fixed for this PR.

packages/client/src/Dashboard.tsx
```diff
@@ -314,6 +315,7 @@ function SearchBar({
           }}
         >
           <InputBase
+            autoFocus={true}
```
hallettj: Oh good - I should have included this line in the first place.

hallettj (approved):
After you fix the failing test this will be ready to merge.

PR #36, opened 2019-07-29
view attachments

carinabquiroz:
While looking at a presentable in a conversation, attachments are displayed with a photo icon and the filename below it, but cannot be selected and opened.


packages/client/src/DisplayContent.tsx
```diff
@@ -92,19 +104,43 @@ function displayMarkdown(text: string, props: object) {
...
 
+function displayAttachment(
+  filename: string | null | undefined,
+  subtype: string,
+  props: object
+) {
+  const name = filename ? (
+    <div {...props}>{filename}</div>
+  ) : (
+    <div {...props}>[Attachment.{subtype}}</div>
```
breitman: Very very small fix, but you have a `}` closing the string instead of matching the `[` at the beginning.

breitman (commented):
Looks good to me; I just suggested the tiniest change!

packages/client/src/Conversation.test.tsx
```diff
@@ -18,6 +19,21 @@ it("displays a conversation", async () => {
   expect(app).toIncludeText("Hello from test")
 })
 
+it("displays an attachment", async () => {
```
hallettj, line 24: Tests! Just what I like to see!

packages/client/src/DisplayContent.tsx
```diff
@@ -92,19 +104,43 @@ function displayMarkdown(text: string, props: object) {
...
 
+function displayAttachment(
+  filename: string | null | undefined,
+  subtype: string,
+  props: object
+) {
+  const name = filename ? (
+    <div {...props}>{filename}</div>
+  ) : (
+    <div {...props}>[Attachment.{subtype}]</div>
```
hallettj, line 115: This seems sensible.

packages/client/src/DisplayContent.tsx
```diff
@@ -40,17 +46,23 @@ export default function DisplayContent({
             handleLink(accountId, navigate, event)
           }
         }
-        if (type === "text" && subtype === "html") {
+        if (disposition === "attachment") {
```
hallettj, line 49: This works for now. In a follow-up feature I would like to design UI for an attachments area at the bottom of the card where the presentable is displayed, or UI elements at the bottom of the card that visually set off attachments. As it stands it is possible that messages will include attachment content before inline content in the MIME tree in which case inline content would not consistently appear before attachment content.

packages/main/src/models/Message.ts
```diff
@@ -45,7 +52,7 @@ export function allContentParts(
   const f = (parts: List<imap.ImapMessagePart>, part: imap.ImapMessagePart) =>
     parts.push(part)
   const zero = List()
-  return foldParts(f, zero, selectAllContent(), struct)
+  return foldPrimaryContent(f, zero, () => true, struct)
```
hallettj: IIRC this change was not necessary for this feature. Would you change this line back?

packages/main/src/models/conversation.ts
```diff
@@ -97,7 +102,9 @@ export function getPresentableElements({
     revision: walkGraph(edits, List([resource]), resource).lastNonConflict
   }))
   return editedContents
-    .groupBy(({ resource }) => resource.message)
+    .groupBy(({ resource, revision }) => {
+      return resource.message
+    })
```
hallettj: This could be changed back since the `revision` parameter is not used.

packages/main/src/models/conversation.ts
```diff
@@ -254,28 +263,22 @@ function getPresentableContent({
...
-      }
-    : fallbackContent()
+  const disposition = part.disposition_type || ""
   return {
-    ...contentMeta,
+    type: part.type || "text",
+    subtype: part.subtype || "plain",
+    content: decoded ? decoded.toString("utf8") : null,
+    disposition:
+      disposition.toLowerCase() === "attachment"
```
hallettj, line 281: The case-insensitive comparison is a good idea.

packages/main/src/resolvers/conversation.ts
```diff
@@ -66,7 +66,8 @@ export const Conversation: ConversationResolvers = {
     const presentables = C.getPresentableElements(conversation)
     const latest = presentables.last(null)
     const content = latest && Seq(latest.contents).first(null)
-    const visible = content && replyParser(content.content, true)
+    const visible =
+      content && replyParser(content.content ? content.content : "", true)
```
hallettj: I could have used a more descriptive variable name to make the intent clear. `replyParser` removes stuff like quoted replies from message content. If there is no content, then there is no point in invoking `replyParser` with an empty string.

In this case if `content.content` is `null` or `undefined` then I would like this entire function to return `null` (or `undefined`).

hallettj (changes requested):
This is great work!

packages/main/src/models/conversation.ts
carinabquiroz: Yeah, gmail sends the disposition_type as all caps.

hallettj (approved):
Looks good!

hallettj:
I merged the search feature which made a bunch of changes to `operations.graphql`. In particular the changes update most of the queries there to use fragments so what we do not have to list the same set of conversation fields over and over for each query. You will need to resolve conflicts in that file, and regenerate `graphl.tsx`.

breitman (approved):

PR #39, opened 2019-08-01
auto-create a draft when writing a reply

breitman:
Was not sure how to test this on front-end, so I stashed my implementation for that for now...


packages/client/src/compose/ReplyForm.tsx
```diff
@@ -20,16 +21,28 @@ type Props = React.FormHTMLAttributes<HTMLFormElement> & {
...
   accountId,
   conversationId,
   replyRecipients,
+  replyDraft,
   ...rest
 }: Props) {
+  const content =
+    replyDraft &&
+    replyDraft.presentables &&
+    replyDraft.presentables[0].contents[0].content
```
hallettj: In theory `presentables[0]` or `contents[0]` could be `undefined`. As I said, you can install `lodash-es` and `@types/lodash-es` and do this:

```ts
import getIn from "lodash-es/get"

// ...

const content = getIn(replyDraft, ["presentables", 0, "contents", 0])
```

packages/client/src/documents/operations.graphql
```diff
@@ -163,6 +163,51 @@ mutation edit(
   }
 }
 
+mutation saveDraft(
+  $accountId: ID!
+  $conversationId: ID!
+  $content: ContentInput!
+) {
+  conversations {
```
hallettj, line 171: Instead of listing fields here please use the `ConversationFieldsForConversationView` fragment.

packages/client/src/documents/operations.graphql
```diff
@@ -235,6 +281,50 @@ fragment ConversationFieldsForConversationView on Conversation {
...
+    id
+    date
+    messageId
+    subject
+    from {
+      name
+      mailbox
+      host
+    }
+    presentables {
```
hallettj: Let's factor out a shared fragment for Presentable fields.

packages/main/schema/schema.graphql
```diff
@@ -64,6 +66,7 @@ type Message {
   messageId: ID!
   subject: String
   from: [Address!]!
+  presentables: [Presentable!]!
```
hallettj: On the Conversation type this field is called `presentableElements`. I do think that `presentables` is a nicer name. I want to change one or the other so that this concept is named consistently.

packages/main/schema/schema.graphql
```diff
@@ -100,6 +103,7 @@ type Presentable {
   from: Address!
   editedAt: String
   editedBy: Address
+  isDraft: Boolean!
```
hallettj, line 106: makes sense

packages/main/src/models/conversation.ts
```diff
@@ -117,6 +119,9 @@ export function getPresentableElements({
         isRead: cache
           .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
           .includes("\\Seen"),
+        isDraft: cache
+          .getFlags(latestEdit ? latestEdit.revision.message.id : message.id)
+          .includes("\\Draft"),
```
hallettj, line 124: Ok; checking the latest revision should work.

packages/main/src/models/conversation.ts
```diff
@@ -153,6 +158,15 @@ function getContentParts(message: cache.Message): List<Revision> {
   })
 }
 
+export function getReplyDraft(messages: cache.Message[]): cache.Message | null {
+  const message = Collection.Indexed(
```
hallettj: Usually one uses a more specific constructor that is a subtype of `Collection.Indexed`, such as `List`.

packages/main/src/models/conversation.ts
```diff
@@ -153,6 +158,15 @@ function getContentParts(message: cache.Message): List<Revision> {
...
+    messages.filter(message => cache.getFlags(message.id).includes("\\Draft"))
+  )
+    .sortBy(message => Number(message.date))
+    .last(null)
```
hallettj, line 166: Is there a test to verify that this gets the most recent draft?

packages/main/src/queue/index.ts
```diff
@@ -249,6 +249,82 @@ const handlers = {
...
 
+  saveDraft: handler({
+    enqueue({
+      accountId,
+      message,
+      box
+    }: {
+      accountId: ID
+      message: ComposedMessage
+      box: { attribute: string }
```
hallettj: The type for a box specifier should be `BoxSpecifier`.

packages/main/src/queue/index.ts
```diff
@@ -249,6 +249,82 @@ const handlers = {
...
+            cache.getBody(messageId, { partID }) || undefined
+          const mimeNode = serialize({
+            attributes,
+            headers,
+            partHeaders,
+            bodies
+          })
+          const raw = await promises.lift1<Buffer>(cb => mimeNode.build(cb))
+          return connectionManager
+            .request(request.actions.append(["\\Draft"], raw, box))
```
hallettj, line 323: good stuff!

packages/main/src/queue/index.ts
```diff
@@ -249,6 +249,82 @@ const handlers = {
...
+      box
+    }: {
+      accountId: ID
+      message: ComposedMessage
+      box: { attribute: string }
+    }) {
+      const { attributes, headers, partHeaders, bodies } = message
+      let messageId: cache.ID | null = null
+
+      db.transaction(() => {
```
hallettj, line 267: This is very similar to the code from the `sendMessage` handler. I'd like to move this to a shared function in `cache/`.

packages/main/src/request/index.ts
```diff
@@ -72,6 +72,20 @@ export const { actions, perform } = combineHandlers({
     })
   },
 
+  append(
+    connection: Connection,
+    flags: string[],
+    msgData: Buffer,
+    box: BoxSpecifier
+  ): R<string> {
```
hallettj: The result is a uid right? Instead of `string`, please use the type `imap.UID`. It is just an alias for `string`, but it helps anyone reading the code to understand what is represented by the result string.

packages/main/src/request/index.ts
```diff
@@ -72,6 +72,20 @@ export const { actions, perform } = combineHandlers({
     })
   },
 
+  append(
+    connection: Connection,
+    flags: string[],
+    msgData: Buffer,
+    box: BoxSpecifier
```
hallettj: For consistency with other handlers please put the `box` argument order second in the list, after `connection`.

packages/main/src/request/index.ts
```diff
@@ -100,6 +114,20 @@ export const { actions, perform } = combineHandlers({
     })
   },
 
+  addLabels(
```
hallettj, line 117: It looks like we are not using this action. But it looks good anyway.

packages/main/src/resolvers/conversation.test.ts
```diff
@@ -361,6 +361,71 @@ describe("when querying conversations", () => {
     })
   })
 
+  it("saves a draft when writing a reply", async () => {
+    const result = await request(
+      `
+      mutation saveDraft($accountId: ID!, $conversationId: ID!, $content: ContentInput!) {
```
hallettj, line 367: Good test! You can remove fields from the query that you do not check in the `expect`.

packages/main/src/resolvers/conversation.ts
```diff
@@ -22,6 +22,21 @@ export const Conversation: ConversationResolvers = {
     return lastUpdated(conversation)
   },
 
+  replyDraft({ messages }: C.Conversation) {
+    const message = C.getReplyDraft(messages)
+
+    return (
```
hallettj, line 28: good stuff

packages/main/src/resolvers/conversation.ts
```diff
@@ -180,6 +195,19 @@ export const ConversationMutations: ConversationMutationsResolvers = {
...
 
+  async saveDraft(_parent, { accountId, id, content }) {
+    const account = mustGetAccount(accountId)
+    const conversation = C.mustGetConversation(id)
+    schedule(
+      actions.saveDraft({
+        box: { attribute: "\\Drafts" },
+        accountId,
+        message: composeReply({ account, content, conversation })
+      })
```
hallettj, line 206: This is going to create a lot of stored messages. Can we replace the existing draft if there is one?

packages/main/src/resolvers/conversation.ts
```diff
@@ -180,6 +195,19 @@ export const ConversationMutations: ConversationMutationsResolvers = {
     return C.mustGetConversation(id)
   },
 
+  async saveDraft(_parent, { accountId, id, content }) {
+    const account = mustGetAccount(accountId)
+    const conversation = C.mustGetConversation(id)
+    schedule(
+      actions.saveDraft({
+        box: { attribute: "\\Drafts" },
```
hallettj, line 203: Yep, this box specifier is good.

packages/main/src/sync.ts
```diff
@@ -337,20 +338,25 @@ class BoxSync {
 }
 
 function matchesCachePolicy(message: imap.ImapMessageAttributes): boolean {
+  let labelMatches = true
```
hallettj: Please declare these variables with `const`. You can collapse the logic below into single RHS expressions.

packages/main/src/sync.ts
```diff
@@ -337,20 +338,25 @@ class BoxSync {
...
 
   if (cachePolicy.since) {
-    if (message.date < cachePolicy.since) {
-      return false
-    }
+    dateMatches = message.date < cachePolicy.since
   }
 
-  return true
+  return labelMatches || flagMatches || dateMatches
```
hallettj: I want this to be `(labelMatches || flagMatches) && dateMatches`.

hallettj (changes requested):

carinabquiroz (approved):

packages/main/src/sync.ts
```diff
@@ -338,25 +338,18 @@ class BoxSync {
 }
 
 function matchesCachePolicy(message: imap.ImapMessageAttributes): boolean {
-  let labelMatches = true
-  let flagMatches = true
-  let dateMatches = true
+  const labels = message["x-gm-labels"] || []
+  const labelMatches =
+    cachePolicy.labels &&
+    !cachePolicy.labels.some(label => labels.includes(label))
```
hallettj, line 344: Yes, I like this style better.

But now I'm noticing that the logic is inverted. `labelMatches` should be `true` if `cachePolicy.labels` is truthy, and some label in the cache policy is also in `labels`. I think the `!` should be removed from this expression, and from the check for flags.

hallettj (changes requested):

PR #41, opened 2019-08-06
open attachment

carinabquiroz:
Downloads attachment when the attachment link is clicked.


packages/client/src/DisplayContent.tsx
```diff
@@ -3,7 +3,7 @@ import PhotoIcon from "@material-ui/icons/Photo"
 import { Location } from "@reach/router"
 import clsx from "clsx"
 import marked from "marked"
-import { parseMidUri } from "poodle-common/lib/models/uri"
+import { parseMidUri, bodyUri } from "poodle-common/lib/models/uri"
```
hallettj: Eslint tells me that the `bodyUri` import is not used.

packages/client/src/DisplayContent.tsx
```diff
@@ -19,7 +19,11 @@ const useStyles = makeStyles(_theme => ({
     whiteSpace: "pre-wrap"
   },
 
-  attachment: {}
+  attachment: {
+    display: "flex",
+    flexDirection: "row",
+    alignItems: "center"
```
hallettj, line 25: This is nice attention to UI detail.

packages/client/src/DisplayContent.tsx
```diff
@@ -47,7 +52,7 @@ export default function DisplayContent({
           }
         }
         if (disposition === "attachment") {
-          return displayAttachment(filename, subtype, {
+          return displayAttachment(filename, subtype, uri, {
```
hallettj, line 55: There is still an issue where where attachments could be mixed together with inline content parts. I added a card to the "This Week" column in the project board to track work on that for an upcoming PR.

packages/client/src/DisplayContent.tsx
```diff
@@ -107,18 +112,17 @@ function displayMarkdown(text: string, props: object) {
...
   subtype: string,
+  uri: string,
   props: object
 ) {
-  const name = filename ? (
-    <div {...props}>{filename}</div>
-  ) : (
-    <div {...props}>[Attachment.{subtype}]</div>
-  )
+  const name = filename ? filename : `Attachment.${subtype}`
```
hallettj: Because the conditional expression is the same as the true-case expression this could be shortened to

```ts
filename || `Attachment.${subtype}`
```

packages/client/src/testing/fixtures.ts
```diff
@@ -88,7 +89,8 @@ export const conversation: graphql.Conversation = {
           content: "Hello from next test",
           disposition: graphql.Disposition.Inline,
           filename: null,
-          name: null
+          name: null,
+          uri: ""
```
hallettj: It would be handy to put URI values in these fields for future use. We have a message cache ID, and you can make up a part ID. For example, `"body:1/2"`

packages/common/src/models/uri.ts
```diff
@@ -38,6 +38,25 @@ export function parseMidUri(
   }
 }
 
+export function bodyUri(cacheId: string, partId: string): URI {
+  const cId = idFromHeaderValue(cacheId)
+  const pId = idFromHeaderValue(partId)
```
hallettj: `idFromHeaderValue` is not necessary here. It is a helper that removes angle brackets from around the input value. That is useful when values are coming from email message headers, because some header values are wrapped with angle brackets. But we generate the cache ID internally, and the part ID is assigned by the IMAP service and does not come from an email message header.

packages/common/src/models/uri.ts
```diff
@@ -38,6 +38,25 @@ export function parseMidUri(
...
+  return `body:${encodeURIComponent(cId)}/${encodeURIComponent(pId)}`
+}
+
+const bodyExp = /(body:)([^/]+)(?:\/(.+))?$/
```
hallettj: I think I told you that this expression would either match both the cache ID and part ID, or `matches` would be `null`. But I was mistaken. Because there is a question mark around one of the groups this expression will match if no part ID is given. Because this URI scheme is only useful when a part ID is given I suggest changing this regular expression.

The sub-expression `(?:\/(.+))?` optionally matches a slash and following text. The use of `?:` inside the outer pair of parenthesis makes it a non-capturing group. That is useful because it allows the entirety of the text captured by the group to be optional (due to the `?` at the end). Content matched by a non-capturing group will not be listen in the `matches` array - that array only includes "captured" groups. You can change this portion of the expression to be required by removing the last `?`. In that case it is not necessary to specify a group that includes the slash. You can change the expression to look like this:

```ts
const bodyExp = /body:([^/]+)\/(.+)$/
```

Notice that I also removed the parenthesis from around `body:` because we do not need to capture the scheme to examine its value. If you remove those parenthesis you need to change the indexes used to get the cache ID and part ID from the `matches` array.

packages/main/package.json
```diff
@@ -29,6 +29,7 @@
     "@types/node-fetch": "^2.1.7",
     "@types/node-uuid": "^0.0.28",
     "@types/nodemailer": "^4.6.7",
+    "@types/tmp": "^0.1.0",
```
hallettj: This looks like a leftover addition that could be removed.

packages/main/src/index.ts
```diff
@@ -1,10 +1,13 @@
...
+import { app, BrowserWindow, ipcMain, protocol } from "electron"
 import contextMenu from "electron-context-menu"
 import isDev from "electron-is-dev"
 import { createIpcExecutor, createSchemaLink } from "graphql-transport-electron"
 import schema from "./schema"
-
+import * as cache from "./cache"
+import { parseBodyUri } from "poodle-common/lib/models/uri"
+import { contentType, filename } from "./models/MessagePart"
+import { PassThrough } from "stream"
```
hallettj, line 7: By the way, I like to keep imports sorted in a certain way. The atom-typescript plugin adds the `typescript:organize-imports` command that sorts imports automatically, and removes unused imports. You can run it by bringing up the command palette with Cmd+Shift+P, and selecting the "Organize module imports" command.

packages/main/src/index.ts
```diff
@@ -37,6 +40,53 @@ function createWindow() {
...
+  protocol.registerStreamProtocol("body", async (request, callback) => {
+    try {
+      const parsed = parseBodyUri(request.url)
+      if (!parsed) {
+        throw new Error(
+          `Unable to parse messageId and partId from URI: ${request.url}`
+        )
+      }
+      const messageId = parsed.messageId
+      const partId = parsed.partId
```
hallettj: If you want you combine this sort of assignment into one statement using destructuring:

```ts
const { messageId, partId } = parsed
```

packages/main/src/index.ts
```diff
@@ -37,6 +40,53 @@ function createWindow() {
...
+    }
+  })
+}
+
+function createStream(input: string | Buffer | null): PassThrough {
+  const stream = new PassThrough()
+  stream.pause()
+  stream.push(input)
+  stream.push(null)
+  return stream
```
hallettj, line 94: Good stuff. I'd be interested to learn more about how this works.

packages/main/src/index.ts
```diff
@@ -37,6 +40,53 @@ function createWindow() {
...
+        statusCode: 200,
+        headers: {
+          "content-type": type,
+          "content-disposition": `attachment; filename= ${file}`
+        },
+        data: stream
+      })
+      stream.resume()
+    } catch (err) {
+      console.error("error serving part content:", err)
```
hallettj, line 77: Logging the error is a good idea.

packages/main/src/index.ts
```diff
@@ -37,6 +40,53 @@ function createWindow() {
...
+      callback({
+        statusCode: 500,
+        headers: {
+          "content-type": "text/plain; charset=utf8"
+        },
+        data: createStream(err.message)
```
hallettj, line 83: It's not necessary for this feature, but it could be a useful exercise to send specific error codes for specific problems. That would be `404 Not Found` if `imapPart` is not defined, `400 Bad Request` if `parsed` is not defined, and `500` if we catch any other exception.

hallettj (changes requested):
This is great! I can save attachments, and the UI looks quite nice. This is a critical feature for the MVP that I have been putting off.

I want you make the change to `parseBodyUri` that I commented on to disallow URIs that do not specify a part ID, remove the `@types/tmp` package, and remove the unused `bodyUri` import. The rest of the feedback is optional or FYI stuff.

hallettj (changes requested):
Looks good! Just one thing: I noticed that you checked in a `package-lock.json` file, which is the lock file that `npm` uses. This project uses yarn and its `yarn.lock` format - we generally want to use `yarn` for everything instead of `npm`. Please remove `package-lock.json`.

PR #42, opened 2019-08-07
Upgrade Electron to v6

hallettj:
This change also installs Electron Forge, which we can use to build installable packages.


package.json
```diff
@@ -1,14 +1,19 @@
...
-    "packages/main"
-  ],
+  "workspaces": {
+    "packages": [
+      "packages/client",
+      "packages/common",
+      "packages/main"
+    ],
+    "nohoist": [
+      "**/*"
```
hallettj, line 11: When using yarn workspaces normally dependencies are installed in the `node_modules/` directory in the project root, and all of the packages in the repository look for dependencies there. Unfortunately Electron Forge (which is run from the main package) is not set up to search parent directories for `node_modules/`. The `nohoist` configuration here instructs yarn to put all dependencies in a `node_modules/` directory in the respective package directory.

package.json
```diff
@@ -1,14 +1,19 @@
...
+    ]
+  },
   "scripts": {
     "diagram": "unflatten -l 3 assets/architecture.dot | dot -Tsvg -o assets/architecture.svg",
-    "postinstall": "scripts/postinstall",
+    "prepare": "scripts/prepare",
```
hallettj, line 16: I changed some scripts from `postinstall` to `prepare` because `prepare` runs after installing dependencies, but does not run when the package itself is installed as a dependency of another package. `postinstall` runs in both cases.

packages/client/package.json
```diff
@@ -10,45 +10,35 @@
...
-    "@types/graphql": "^14.2.2",
-    "@types/is-url": "^1.2.28",
-    "@types/marked": "^0.6.5",
-    "@types/node": "11.11.3",
-    "@types/object.pick": "^1.3.0",
-    "@types/reach__router": "^1.2.4",
-    "@types/react": "^16.8.22",
-    "@types/react-autosuggest": "^9.3.8",
-    "@types/react-dom": "^16.8.4",
-    "@types/string-hash": "^1.1.1",
```
hallettj, line 39: I moved `@types/*` packages to dev dependencies because they are only useful during type checking, and do not need to be packaged with the final build of the app.

packages/client/package.json
```diff
@@ -10,45 +10,35 @@
...
     "apollo-client": "^2.6.2",
     "autosuggest-highlight": "^3.1.1",
     "clsx": "^1.0.4",
-    "electron": "^4.1.0",
     "email-addresses": "^3.0.3",
     "graphql": "^14.4.2",
     "graphql-tag": "^2.10.1",
     "graphql-transport-electron": "^1.0.1",
-    "immutable": "^4.0.0-rc.12",
+    "immutable": "^3.8.2",
```
hallettj, line 38: I changed the dependency version for `immutable` in client to match the version that the `slate` package depends on.

packages/client/package.json
```diff
@@ -10,45 +10,35 @@
...
     "is-url": "^1.2.4",
     "marked": "^0.6.2",
+    "moment": "^2.24.0",
```
hallettj, line 41: Now that dependencies are not hoisted anymore there were errors importing packages that were listed as dependencies in one of the other packages. Client depends on `moment`, and now that needs to be made explicit by listing the dependency in `package.json`.

carinabquiroz (approved):

PR #43, opened 2019-08-09
Build distributable packages

hallettj:
This change adds new build scripts in the main package. `yarn package` builds a production version of the app that can be run directly by running `packages/main/out/poodle-$TARGET/poodle`.

Running `yarn make` builds installers. In theory we are set up to build installable packages for Linux, MacOS, and Windows. Packages are output to `packages/main/out/make/`. I have only tested the Linux `.deb` package so far. My next step is to set up automated builds in CI, automated publishing, and hopefully we'll be able to set up automatic updates for end users.


packages/client/config-overrides.js
```diff
@@ -0,0 +1,40 @@
...
+  "default-src": ["'none'"],
+  "script-src": [],
+  "style-src": ["'self'", "'unsafe-inline'"],
+  "img-src": ["'self'", "data:", "cid:", "mid:"],
+  "child-src": ["data:"],
+  "font-src": ["'self'"],
+  "media-src": ["'self'", "data:", "cid:", "mid:"],
+  "object-src": ["'none'"],
+  "form-action": ["'none'"]
+}
```
hallettj, line 14: The packaged version of the app loads html, css, and javascript assets via the `file:` protocol instead of `http:`. Browsers treat these protocols differently in a few ways. In particular using a content security policy such as `'self'` does not work well with `file:` assets. We need to switch to a more secure nonce policy, where the browser will only execute a script tag if it has a `nonce` attribute with a value that is listed in the content security policy. Script tags are written by Webpack, so we need to modify the Webpack configuration to include nonces; and because the client package uses create-react-app, we need to use tools such as `customize-cra` to patch create-react-app's Webpack configuration.

packages/client/config-overrides.js
```diff
@@ -0,0 +1,40 @@
...
+
+const cspConfigPolicy_dev = {
+  ...cspConfigPolicy,
+  "connect-src": ["'self'", "ws://localhost:3000"],
```
hallettj, line 18: Create-react-app uses a websocket connection to automatically reload the browser when code changes are made. We want a content security policy (CSP) in development that allows this. But we want a more locked-down configuration in production.

packages/client/config-overrides.js
```diff
@@ -0,0 +1,40 @@
...
+  "script-src": ["chrome-extension:"]
```
hallettj, line 19: We use some Chrome extensions for development to introspect React and Apollo state. We need to allow the `chrome-extension:` scheme in the CSP to allow these to load. But we don't need to allow this in production.

packages/client/package.json
```diff
@@ -5,14 +5,15 @@
   "description": "Experimental email client with social features",
   "author": "Jesse Hallett <jesse@sitr.us>",
   "license": "Apache-2.0",
+  "homepage": "./",
```
hallettj, line 8: This `homepage` setting is how you configure create-react-app to serve assets from a non-root host location.

packages/client/package.json
```diff
@@ -5,14 +5,15 @@
...
   "scripts": {
     "prestart": "cd ../common && yarn build",
-    "start": "cross-env BROWSER=none react-scripts start",
+    "start": "cross-env BROWSER=none react-app-rewired start",
     "prebuild": "cd ../common && yarn build",
-    "build": "react-scripts build",
+    "build": "cross-env INLINE_RUNTIME_CHUNK=false react-app-rewired build",
```
hallettj, line 13: Setting `INLINE_RUNTIME_CHUNK` configures create-react-app to use a non-inline script tag to load bootstrapping code. This is necessary so that we can put a nonce on that script tag.

packages/client/public/index.html
```diff
@@ -2,19 +2,7 @@
...
-        connect-src 'self' ws://localhost:3000;
-        script-src 'self' chrome-extension:;
-        style-src 'self' 'unsafe-inline' http://fonts.googleapis.com;
-        img-src 'self' data: mid:;
-        child-src 'self' data:;
-        font-src 'self' http://fonts.gstatic.com;
-        media-src 'self';
-        form-action 'none';
-        " />
+    <meta http-equiv="Content-Security-Policy" content="" />
```
hallettj, line 5: The CSP is now generated by Webpack, not written by hand. We put in a `meta` tag with an empty `content` value to show Webpack where to emit the policy.

packages/main/package.json
```diff
@@ -89,6 +69,7 @@
     "@electron-forge/maker-rpm": "^6.0.0-beta.43",
     "@electron-forge/maker-squirrel": "^6.0.0-beta.43",
     "@electron-forge/maker-zip": "^6.0.0-beta.43",
+    "@electron-forge/plugin-auto-unpack-natives": "^6.0.0-beta.43",
```
hallettj, line 72: Electron can load assets from a single combined file called an "asar". It is similar to a `.tar` file - many files are combined into one. The asar includes the apps javascript files for both the frontend and backend, node modules, html and css assets, etc. But apparently loading dependencies that use native code from an asar is inefficient. This plugin excludes just those dependencies from the asar, leaving them as separate files.

packages/main/src/db.ts
```diff
@@ -1,45 +1,48 @@
...
-function getDbPath() {
+export function getDbPath(dbName: string, isDev: boolean): string {
   const cacheDir = xdgBasedir.cache
   if (!cacheDir) {
     throw new Error(
       "Could not locate XDG cache location! Are you running without a home directory?"
     )
   }
-  return path.join(cacheDir, "poodle", "db.sqlite")
+  return path.join(cacheDir, isDev ? "poodle-dev" : "poodle", dbName)
```
hallettj, line 45: I made a change so that the app uses a different cache location for development mode vs production. That will allow us to use the installed version of Poodle for day-to-day email tasks, while developing and testing the app without interfering with the production databases.

packages/main/src/index.ts
```diff
@@ -8,11 +8,18 @@ import { PassThrough } from "stream"
...
 
 // TODO: We're having an issue checking the TLS certificate for Google's IMAP
 // service
 process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
 
+// TODO: Is there a better way to either load frontend assets from the asar, or
+// to reference unpacked assets?
+const appUrl = isDev
+  ? "http://localhost:3000"
+  : "file://" + path.join(__dirname, "..", "public", "index.html")
```
hallettj, line 19: Thanks to some Electron magic we can reference files that are actually packaged inside the asar as though they are ordinary files in the host filesystem.

carinabquiroz (approved):

PR #44, opened 2019-08-12
add attachment

carinabquiroz:
User can click on the add attachment button and add attachments to both the compose and reply form. Can also click on the "X" next to an attachment to remove it.


packages/client/src/compose/Compose.tsx
```diff
@@ -107,6 +127,12 @@ export default function Compose({ accountId }: Props) {
     } catch (error) {}
   }
 
+  function handleAttachments(event: React.FormEvent<HTMLInputElement>) {
+    const files = Array.from(event.currentTarget.files || [])
+    const updatedFiles = [...attachments, ...files]
+    setAttachments(updatedFiles)
```
hallettj: This is looking pretty nice.

packages/client/src/compose/Compose.tsx
```diff
@@ -167,11 +193,39 @@ export default function Compose({ accountId }: Props) {
             placeholder="Write your message here."
             value={content}
           />
-          <Button color="primary" variant="contained" type="submit">
-            Send
-          </Button>
+          {Attachments(attachments, classes.attachment)}
+          <span className={classes.actionRow}>
```
hallettj: Good stuff

packages/client/src/compose/Compose.tsx
```diff
@@ -167,11 +193,39 @@ export default function Compose({ accountId }: Props) {
...
+
+function Attachments(attachments: File[], style: string) {
+  return attachments.map(attachment => {
+    return (
+      <div className={style}>
+        <PhotoIcon />
+        {attachment.name}
+      </div>
+    )
+  })
```
hallettj: It would be more idiomatic to make `Attachments` a React component. You can create a React component that effectively returns an array of elements by wrapping the array in a fragment like this,

```ts
function Attachments({ attachments, style }: { attachments: File[], style: string }) {
  const elements = attachments.map(attachment => {
    return (
      <div className={style}>
        <PhotoIcon />
        {attachment.name}
     </div>
   )
  return <>{elements}</>
})
```

The fragment (`<></>`) is treated as a single element for purposes of the React Element tree, but is not mapped to a DOM element. Instead the fragment's children are all put into the DOM tree in the fragment's place.

Also, the `div` element needs a `key` property. You could use a combination of the file name and array index as a key - I think it is possible that there could be multiple attachments with the same file name.

packages/client/src/compose/ReplyForm.tsx
```diff
@@ -79,13 +110,39 @@ export default function ReplyForm({
...
+            type="file"
+            onChange={handleAttachments}
+          />
+          <label htmlFor="add-attachment-button">
+            <Tooltip title={"Add Attachment"}>
+              <IconButton component="span">
+                <AttachFileIcon />
+              </IconButton>
+            </Tooltip>
+          </label>
```
hallettj: As you mentioned, we want to create reusable components for this stuff. We can work together on implementing the compound components pattern.

hallettj (changes requested):

hallettj (approved):
Looks great!

